PR #54, opened 2015-03-24
Add generate_sync method to tarantool.Connection class

shveenkov:
These improvements will not affect the implementation of the current driver code.

However, this would greatly simplify the implementation of gtarantool driver https://github.com/shveenkov/gtarantool.

GRequest, GResponse classes will be removed, also, do not need to copy-paste methods select / insert / delete in TarantoolCoroConnection.

### DIFF
--- a/tarantool/__init__.py
+++ b/tarantool/__init__.py
@@ -24,7 +24,7 @@ from tarantool.schema import (
 )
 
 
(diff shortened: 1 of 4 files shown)


dreadatour:
+1

bigbes:
Merged

PR #66, opened 2016-01-28
This statement without a specific error code (only index) leads to Ke…

mkaplenko:
…yError Exception in 258 line, when attempt to get error code with zero key mapping in the linux ubuntu system (or others)..


bigbes:
Thank you!

Issue #95, opened 2017-11-13
Error when connecting to tarantool


skythet:
Trying to connect to tarantool: `tarantool.connect('127.0.0.1', 3301)`. But the following error occurs:

```
Traceback (most recent call last):
  File "server.py", line 2, in <module>
    from file_server.application import application
  File "file_server/application.py", line 10, in <module>
    tarantool_connection = tarantool.connect(host, port)
  File "ENV/lib/python3.5/site-packages/tarantool/__init__.py", line 50, in connect
    encoding=encoding)
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 127, in __init__
    self.connect()
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 203, in connect
    self.load_schema()
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 344, in load_schema
    self.schema.fetch_space_all()
  File "ENV/lib/python3.5/site-packages/tarantool/schema.py", line 130, in fetch_space_all
    space_rows = self.fetch_space_from(None)
  File "ENV/lib/python3.5/site-packages/tarantool/schema.py", line 117, in fetch_space_from
    index=_index)
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 759, in select
    response = self._send_request(request)
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 341, in _send_request
    return self._send_request_wo_reconnect(request)
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 264, in _send_request_wo_reconnect
    self.update_schema(e.schema_version)
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 349, in update_schema
    self.flush_schema()
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 353, in flush_schema
    self.load_schema()
  File "ENV/lib/python3.5/site-packages/tarantool/connection.py", line 345, in load_schema
    self.schema.fetch_index_all()
  File "ENV/lib/python3.5/site-packages/tarantool/schema.py", line 166, in fetch_index_all
    SchemaIndex(row, self.schema[row[0]])
  File "ENV/lib/python3.5/site-packages/tarantool/schema.py", line 29, in __init__
    for k, v in index_row[5]:
ValueError: too many values to unpack (expected 2)
```
Python version: `Python 3.5.2`
Python tarantool connector version:
```
pip show tarantool
[...]

PR #96, opened 2017-11-13
support for new format indexes

skythet:
Try to fix issue #95

### DIFF
--- a/tarantool/connection.py
+++ b/tarantool/connection.py
@@ -84,6 +84,7 @@ class Connection(object):
     def __init__(self, host, port,
                  user=None,
                  password=None,
+                 socket=None,
                  socket_timeout=SOCKET_TIMEOUT,
                  reconnect_max_attempts=RECONNECT_MAX_ATTEMPTS,
                  reconnect_delay=RECONNECT_DELAY,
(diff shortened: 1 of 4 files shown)


tarantool/connection.py
```diff
@@ -134,7 +136,7 @@ def close(self):
         self._socket = None
 
     def connect_basic(self):
-        if self.host == None:
+        if self.socket:
```
bigbes: Why're you doing this incompatible changes? use separate PR's for them.

tarantool/schema.py
```diff
@@ -26,8 +26,11 @@ def __init__(self, index_row, space):
         self.unique = index_row[4]
         self.parts = []
         if isinstance(index_row[5], (list, tuple)):
-            for k, v in index_row[5]:
-                self.parts.append((k, v))
+            for part in range(len(index_row[5])):
```
bigbes: Please, add tests for this part of code

tarantool/connection.py
skythet: Sorry, I'll send another PR for this. I thought this repo no longer active

tarantool/schema.py
skythet: OK. But, I've changed format of parts to dict `{'field': part[0], 'type': part[1]}`, because for me parts doesn't matter. I don't use them. But this change brokes backward compatibility. So, I should do this by older way? I mean like this:
```
if isinstance(part, list):
    self.parts.append((part[0], part[1]))
else:
    self.parts.append((part['field'], part['type']))
```

skythet:
I created another PR with tests. Because changes in this PR not relevant.
This PR can be closed.

PR #99, opened 2018-02-06
Socket connection support

skythet:
`tarantool.connect(socket="/var/lib/tarantool/tarantool.sock")`


alg1973:
connect(host=None, port="/var/lib/tarantool/tarantool.sock") is intended for that you've done. However I like your idea. It's good to have explicit unix socket API parameters/functions.

Issue #106, opened 2018-08-30
Reconnection strategy - round-robin between instances


sharonovd:
Use-case: tarantool is deployed as a geo-distributed cluster, and when one DC goes down, application must switch to reserve DC with minimal service disruption.

Proposal 1: add built-in reconnect strategy which would take several connection URI's as a config, and would quickly to establish connection to them in a circle, returning the first succesful one.
Edge-case: if all URIs are unavailable, return error.

Proposal 2: implement overridable connection strategy class (as in tarantool-java), and provide proposal 1 as example.

PR #107, opened 2018-09-13
inconnect bug fix

kbelyavs:
Add missing inconnect = false.
This bug leads to only one disconnection allow.

Needed for #106


bigbes:
As quick-fix: LGTM

PR #109, opened 2018-09-19
Implement reconnection strategy class

### LINKED ISSUE #106: Reconnection strategy - round-robin between instances

kbelyavs:
Extend base Connection class to support a list of nodes and an
optional Strategy parameter to choose next item from this list.

Add built-in reconnect strategy class based on Round-Robin alg.
@params:
 - peers

Return next connection or an error if all URIs are unavailable.

Closes #106


client.py
```diff
@@ -0,0 +1,85 @@
...
+    '''
+
+    return MeshConnection(addrs=addrs,
+                          user=user,
+                          password=password,
+                          connect_now=connect_now,
+                          encoding=encoding,
+                          Strategy=Strategy)
+
+addrs=(
```
Totktonada: I think we should not place some hardcoded actions like these ones (here and below) in the module.

Totktonada: Esp. when they can fail because of external conditionals (whether are tarantools running on these ports).

client.py
```diff
@@ -0,0 +1,85 @@
...
+
+import sys
+import time
+
+from tarantool.connection import Connection
+from tarantool.error import NetworkError
+from tarantool.utils import ENCODING_DEFAULT
+
+class RoundRobinStrategy(object):
+    def __init__(self, addrs):
```
Totktonada: The course code uses `addrs`, but the commit message uses `peers` term.

client.py
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
```
Totktonada: 1. Why not in tarantool/ directory?
2. We need to ensure the code works with python2 as well as python3.
3. flake8 reports some warings.

client.py
```diff
@@ -0,0 +1,85 @@
...
+# -*- coding: utf-8 -*-
```
Totktonada: We discussed it verbally, but I formalize here a request for writing some test for this functionality.

Totktonada:
```
$ flake8 tarantool/__init__.py tarantool/mesh_connection.py 
tarantool/__init__.py:53:1: E302 expected 2 blank lines, found 1
tarantool/mesh_connection.py:16:1: E302 expected 2 blank lines, found 1
tarantool/mesh_connection.py:26:1: E302 expected 2 blank lines, found 1
tarantool/mesh_connection.py:38:9: F841 local variable 'addr' is assigned to but never used
tarantool/mesh_connection.py:45:33: E128 continuation line under-indented for visual indent
tarantool/mesh_connection.py:46:33: E128 continuation line under-indented for visual indent
tarantool/mesh_connection.py:47:33: E128 continuation line under-indented for visual indent
```

test/cluster-py/multi.test.py
```diff
@@ -0,0 +1,116 @@
...
+    SOCKET_TIMEOUT,
+    RECONNECT_DELAY,
+)
+from tarantool.error import NetworkError
+from tarantool.utils import ENCODING_DEFAULT
+
+REPLICA_N = 3
+ROW_N = REPLICA_N * 20
+
+def check_cluster(addrs):
```
Totktonada: I think the test can be written in much easier manner. You don't need to create cluster and setup the replication, just start two instances with a bit different data. Threads and sleeps don't necessary too, the test can be written in sequencial manner (insert data to servers1, inerst data to server2, create MeshConnection, make select/call, down instance 1, make select/call, up instance 1, make select/call, down instance 2, make select/call, down instance 1, make select/call and receive an error).

test/cluster-py/replica.lua
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env tarantool
+box_cfg_done = false
```
Totktonada: Is not used.

test/cluster-py/suite.ini
```diff
@@ -0,0 +1,5 @@
+[default]
+core = tarantool
+script =  master.lua
+description = tarantool/box, replication
```
Totktonada: Description seems to be not about the suite we have here.

test/cluster-py/multi.test.py
```diff
@@ -0,0 +1,77 @@
...
+sources = []
+for server in cluster[1:]:
+    sources.append(yaml.load(server.admin('box.cfg.listen', silent=True))[0])
+
+addrs = []
+for addr in sources:
+    host, port = addr.split(':')
+    addrs.append({'host': host, 'port': int(port)})
+
+thread = Thread(target=check_cluster, args=(addrs, ))
```
Totktonada: I think it is better to create purely sequencial test and don't lean on timings. Like so (schematically):

```
setup instances
setup mesh connection
perform select
stop instance 1
perform select
stop instance 2
start instance 1
perform select
stop instance 1
perform select
```

test/cluster-py/multi.test.py
```diff
@@ -0,0 +1,77 @@
...
+            print 'NetworkError !'
+            break
+        except Exception as e:
+            print e
+        time.sleep(1)
+
+
+# Start replicas
+master = server
+cluster = [master]
```
Totktonada: We can use the default instance and one more in the test. No need to start 'dump' default server, I think.

test/cluster-py/multi.test.py
```diff
@@ -0,0 +1,77 @@
...
+sys.path.append('../tarantool')
+from mesh_connection import MeshConnection
+from tarantool.const import (
+    SOCKET_TIMEOUT,
+    RECONNECT_DELAY,
+)
+from tarantool.error import NetworkError
+from tarantool.utils import ENCODING_DEFAULT
+
+REPLICA_N = 2
```
Totktonada: Nit: It is not a replica, but just separate tarantool instance. Maybe naming should be fixed for other files too.

Makefile
```diff
@@ -1,5 +1,6 @@
+.PHONY: test
 test:
-	python -m pytest
+	python setup.py test
```
Totktonada: I think here we should run unit and functional tests both.

test/cluster-py/replica.lua
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env tarantool
+
+local INSTANCE_ID = string.match(arg[0], "%d")
+local SOCKET_DIR = require('fio').cwd()
+
+local function instance_uri(instance_id)
+    return SOCKET_DIR..'/autobootstrap'..instance_id..'.sock';
```
Totktonada: Why replica? Why autobootstrap? Why multi?

I think we should adjust naming.

test/cluster-py/multi.test.py
```diff
@@ -0,0 +1,72 @@
+import sys
+import os
+from threading import Thread
```
Totktonada: Is not used.

Totktonada:
I'll ok after:

* Handle CI somehow (issue at least).
* Use appropriate naming in the test (replica, multi, autobootstrap → ...).

Also I think we can use default server as the first instance in the test (and setup one more).

Please, proceed with that and merge. I don't think I should look into that again.

PR #140, opened 2019-05-24
Fix deprecation warning

olegrok:
```
/usr/local/lib/python3.7/site-packages/tarantool/response.py:25
  /usr/local/lib/python3.7/site-packages/tarantool/response.py:25: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    class Response(collections.Sequence):
```


Totktonada:
Thanks! Pushed in bce4b88454ba2cc586d41bcd929a55496d6ca070.

PR #142, opened 2019-06-03
Hotfix schema version on auth

### LINKED ISSUE #141: SchemaReloadException can be risen during authentification
We cannot fetch a new schema before successful authentification, so it worth to always authentificate with zero schema id (to suppress server-side schema id check).

robinhood23rus:
No tests here because it appear only on mesh connection with refresh.
closes #141

### DIFF
--- a/tarantool/request.py
+++ b/tarantool/request.py
@@ -134,6 +134,16 @@ class RequestAuthenticate(Request):
         request_body = msgpack.dumps({IPROTO_USER_NAME: user,
                                       IPROTO_TUPLE: ("chap-sha1", scramble)})
         self._body = request_body
+    
+    def header(self, length):
+        self._sync = self.conn.generate_sync()
+        # Set IPROTO_SCHEMA_ID: 0 to avoid SchemaReloadException
+        # It is ok to use 0 in auth every time
+        header = msgpack.dumps({IPROTO_CODE: self.request_type,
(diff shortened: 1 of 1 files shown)


tarantool/connection.py
```diff
@@ -433,6 +433,7 @@ def authenticate(self, user, password):
         if not self._socket:
             return self._opt_reconnect()
 
+        self.schema_version = 0
```
Totktonada: Here we set zero schema id for this and all subsequent requests. Of course, it will work, but refetching the entire schema after each reconnection is not optimal. I would set schema version only for this certain request. As far as I see it can be done with reloading `header()` method in RequestAuthenticate.

robinhood23rus: > refetching the entire schema after each reconnection is not optimal
Updating schema is everywhere =) Overloading header method will do little job. It need to be refactored in handshake, send_request_* and connection. May be somewhere else... check is needed. And also there is case of flush schema on auth fail.

Totktonada: Why the overloading does not do the job? Don't get it.

robinhood23rus: Made schema zero in RequestAuth header. Tested in #135 works as expected.

Totktonada:
Removed trailing spaces, updated the commit message. Pushed in aa8f115026cc565114330b1240e60331a5628295.

PR #145, opened 2019-06-10
Fix response obj string repr in case of no data

### LINKED ISSUE #139: No _data in response in some cases
When calling Connection authenticate there is not data in response. In interactive console exception raised 
`--> 236         for tpl in self._data:
    237             output.extend(("- ", repr(tpl), "\n"))
    238         if len(output) > 0:

TypeError: 'NoneType' object is not iterable`

https://github.com/tarantool/tarantool-python/blob/debe67634bc83d8c1b1687c623629cc861e1c4fc/tarantool/response.py#L236

Possible fix:
for i in r._data or ()

robinhood23rus:
Resolves #139

### DIFF
--- a/tarantool/response.py
+++ b/tarantool/response.py
@@ -238,7 +238,7 @@ class Response(Sequence):
                 }
             }, sort_keys = True, indent = 4)
         output = []
-        for tpl in self._data:
+        for tpl in self._data or ():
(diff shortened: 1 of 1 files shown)


Totktonada:
I'm okay with the code. Updated the commit message with more information.

PR #151, opened 2019-09-06
Fix documentation links in README

RunsFor:


Totktonada:
Please, don't create extra merge commits: they are never useful.

Issue #155, opened 2020-03-12
New python-msgpack version brokes connector


RepentantGopher:
New version of python-msgpack library was released which enforce only string keys in map by default.
This change brokes whole connector because version of python msgpack library is not enforced in requirements.

Commit https://github.com/msgpack/msgpack-python/commit/d8e3cf0563989a660398318a7c788645124e1d8b

Error on connection to tarantool:
```
test_tags.py:6: in <module>
    from .common import (
common.py:22: in <module>
    operational_storage = tarantool.connect("localhost", 3303, "tarantool", "tarantool")
venv/lib/python3.7/site-packages/tarantool/__init__.py:51: in connect
    encoding=encoding)
venv/lib/python3.7/site-packages/tarantool/connection.py:130: in init
    self.connect()
venv/lib/python3.7/site-packages/tarantool/connection.py:208: in connect
    raise NetworkError(e)
E   tarantool.error.NetworkError: int is not allowed for map key
```

pcdinh:
I got the same issue in Python 3.8, Tarantool 0.6.6

Totktonada:
Should be fixed with PR #156.

artembo:
In the new release of msgpack also encoding parameter was removed.
> Remove encoding option from the Packer and Unpacker.

@Totktonada suggested

If the encoding option is given to the connector, then you can programmatically go through recursively and do .decode (encoding = 'utf-8') for all bytes (Python 3) or str (Python 2) values. It seems like this should work, and fully transparently support everything that was previously supported:

Python 2:
```
>>> import msgpack
>>> msgpack.unpackb(b'\xd9\x08\xd0\xb0\xd0\xb1\xd0\xb2\xd0\xb3', raw=True).decode(encoding='utf-8')
u'\u0430\u0431\u0432\u0433'
>>> msgpack.unpackb(b'\xd9\x08\xd0\xb0\xd0\xb1\xd0\xb2\xd0\xb3', encoding='utf-8')
u'\u0430\u0431\u0432\u0433'
```
Python 3:
```
>>> import msgpack
>>> msgpack.unpackb(b'\xd9\x08\xd0\xb0\xd0\xb1\xd0\xb2\xd0\xb3', raw=True).decode(encoding='utf-8')
'абвг'
>>> msgpack.unpackb(b'\xd9\x08\xd0\xb0\xd0\xb1\xd0\xb2\xd0\xb3', encoding='utf-8')
[...]

PR #156, opened 2020-03-12
Support msgpack 1.0.0

pcdinh:
In msgpack 1.0.0, the default value of strict_map_key is changed to True to avoid hashdos. However, it will break the Tarantool driver, causing an exception: 

```
ValueError: int is not allowed for map key
```

As a result, the driver is unable to parse server response. The issue is reported at https://github.com/tarantool/tarantool-python/issues/155

### DIFF
--- a/tarantool/response.py
+++ b/tarantool/response.py
@@ -54,7 +54,7 @@ class Response(Sequence):
             # Get rid of the following warning.
             # > PendingDeprecationWarning: encoding is deprecated,
             # > Use raw=False instead.
-            unpacker = msgpack.Unpacker(use_list=True, raw=False)
+            unpacker = msgpack.Unpacker(use_list=True, raw=False, strict_map_key=False)
         elif conn.encoding is not None:
             unpacker = msgpack.Unpacker(use_list=True, encoding=conn.encoding)
         else:
--- a/tarantool/response.py
+++ b/tarantool/response.py
@@ -54,11 +54,20 @@ class Response(Sequence):
             # Get rid of the following warning.
             # > PendingDeprecationWarning: encoding is deprecated,
             # > Use raw=False instead.
-            unpacker = msgpack.Unpacker(use_list=True, raw=False, strict_map_key=False)
+            if msgpack.version >= (1, 0, 0):
+                unpacker = msgpack.Unpacker(use_list=True, raw=False, strict_map_key=False)
+            else:
+                unpacker = msgpack.Unpacker(use_list=True, raw=False)
         elif conn.encoding is not None:
-            unpacker = msgpack.Unpacker(use_list=True, encoding=conn.encoding)
+            if msgpack.version >= (1, 0, 0):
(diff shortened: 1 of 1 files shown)


Totktonada:
In CI:

> TypeError: __init__() got an unexpected keyword argument 'strict_map_key'

It seems we should support both old and new implementations.

tarantool/response.py
```diff
@@ -54,11 +54,20 @@ def __init__(self, conn, response):
...
+            if msgpack.version >= (1, 0, 0):
+                unpacker = msgpack.Unpacker(use_list=True, encoding=conn.encoding, strict_map_key=False)
+            else:
+                unpacker = msgpack.Unpacker(use_list=True, encoding=conn.encoding)
         else:
-            unpacker = msgpack.Unpacker(use_list=True)
+            if msgpack.version >= (1, 0, 0):
+                unpacker = msgpack.Unpacker(use_list=True, strict_map_key=False)
+            else:
+                unpacker = msgpack.Unpacker(use_list=True)
```
Totktonada, line 70: We can reduce code duplication by unpacking a dictionary to keyword parameters. A kind of this (not tested):

```python
kwargs = dict(use_list=True)
if msgpack.version >= (1, 0, 0):
    # XXX: Explain why it is necessary.
    kwargs['strict_map_key'] = False
if msgpack.version >= (0, 5, 2) and conn.encoding == 'utf-8':
    # Get rid of the following warning.
    # > PendingDeprecationWarning: encoding is deprecated,
    # > Use raw=False instead.
    kwargs['raw'] = False
elif conn.encoding is not None:
    kwargs['encoding'] = conn.encoding

unpacker = msgpack.Unpacker(**kwargs)
```

Totktonada:
From python-msgpack 1.0.0 [release notes](https://github.com/msgpack/msgpack-python/blob/d9ead81021c7b5f034a0475bf9a88e9612cc8e84/ChangeLog.rst):

> * Remove `encoding` option from the Packer and Unpacker.

Are'we affected by this?

Totktonada:
> From python-msgpack 1.0.0 [release notes](https://github.com/msgpack/msgpack-python/blob/d9ead81021c7b5f034a0475bf9a88e9612cc8e84/ChangeLog.rst):
> 
> > * Remove `encoding` option from the Packer and Unpacker.
> 
> Are'we affected by this?

In fact, [no](https://github.com/tarantool/tarantool-python/issues/155#issuecomment-680512896):

> I dived a bit further and it reminds me changes we did for msgpack-0.5.2.
> 
> We already use `raw=False` when `encoding` is `'utf-8'` since msgpack-0.5.2 ([2097884](https://github.com/tarantool/tarantool-python/commit/2097884cc531a419cdf30e1e153094994faf274b)), because otherwise it emits the deprecation warning regarding `encoding` option usage. So we should just continue passing `raw=False` on msgpack-1.0.0.

Totktonada:
Closed in favor of PR #173.

PR #167, opened 2020-08-24
msgpack: Add 1.0.0 version support

artembo:
Fix for msgpack version 1.0.0 where `strict_map_key` param was removed.

Please see #155 issue for details.

Since msgpack 1.0.0 was released Unpacker does not have strict_map_key param
and it is True by default now.

It makes the Connector fail with an error: `ValueError: int is not allowed for map key`

The PR contains a fix suggested by @Totktonada [here](https://github.com/tarantool/tarantool-python/pull/156#discussion_r452565743) in the PR #156 which is stuck.

Close related PR #156 please when this one is merged in master.

### DIFF
--- a/tarantool/response.py
+++ b/tarantool/response.py
@@ -50,16 +50,19 @@ class Response(Sequence):
         # created in the __new__().
         # super(Response, self).__init__()
 
+        kwargs = dict(use_list=True)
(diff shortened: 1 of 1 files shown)


Totktonada:
Closed in favor of PR #173.

Issue #105, opened 2018-07-29
Unpack for binary data


copyhold:
Hi,

My data in tarantool stored as binary strings, they are not UTF in any manner. It's just a sequence of bytes.
When trying to select data with this module , unpacker fails with error:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/dist-packages/tarantool/space.py", line 75, in select
    return self.connection.select(self.space_no, *args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/tarantool/connection.py", line 775, in select
    response = self._send_request(request)
  File "/usr/local/lib/python3.6/dist-packages/tarantool/connection.py", line 357, in _send_request
    return self._send_request_wo_reconnect(request)
  File "/usr/local/lib/python3.6/dist-packages/tarantool/connection.py", line 264, in _send_request_wo_reconnect
    response = Response(self, self._read_response())
  File "/usr/local/lib/python3.6/dist-packages/tarantool/response.py", line 62, in __init__
    self._body = unpacker.unpack()
  File "msgpack/_unpacker.pyx", line 519, in msgpack._unpacker.Unpacker.unpack
  File "msgpack/_unpacker.pyx", line 499, in msgpack._unpacker.Unpacker._unpack
msgpack.exceptions.UnpackValueError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte

```

Can this be resolved somehow?

31hkim:
Any success with fixing that? Still have that problem with Tarantool 2.2 and connector 0.6.6 even when I don't select binary data from a space containing it

Totktonada:
It seems, we cannot just set `encoding` to `None` when creating a connection, because at least schema reload does not support this.

However we can temporary set encoding to `None`:

```diff
diff --git a/unit/suites/test_dml.py b/unit/suites/test_dml.py
index 31e821e..1264180 100644
--- a/unit/suites/test_dml.py
+++ b/unit/suites/test_dml.py
@@ -70,6 +70,12 @@ class TestSuite_Request(unittest.TestCase):
                     self.con.insert('space_1', [i, i%5, 'tuple_'+str(i)])[0],
                     [i, i%5, 'tuple_'+str(i)]
             )
+        saved_encoding = self.con.encoding
+        self.con.encoding = None
+        t = [1001, 1, b'\xff']
+        self.assertEqual(self.con.insert('space_1', t)[0], t)
[...]

PR #172, opened 2020-08-26
schema: support encoding=None connections

Totktonada:
Several different problems are fixed here, but all have the same root.
When a connection encoding is None (it is default on Python 2 and may be
set explicitly on Python 3), all mp_str values are decoded into bytes,
not Unicode strings (note that bytes is alias for str in Python 2). But
the database schema parsing code have assumptions that _vspace / _vindex
values are Unicode strings.

The resolved problems are the following:

1. Default encoding in bytes#decode() method is 'ascii', however names
   in tarantool can contain symbols beyond ASCII symbol table. Set
   'utf-8' for names decoding.
2. Convert all binary values into Unicode strings before parse or store
   them. This allows further correct accesses to the local schema
   representation.
3. Convert binary parameters like space, index or field name into
   Unicode strings, when a schema is accessed to don't trigger redundant
   schema refetching.

Those problems are briefly mentioned in [1].

Tested manually with Python 2 and Python 3: my testing tarantool
instance has a space with name '©' and after the changes I'm able to
connect to it when the connection encoding is set to None. Also I
verified that schema is not fetched each time when I do
\<connection\>.select('©') in Python 2 (where such string literal is str /
bytes, not Unicode string).

Added relevant test cases as separate commits within the PR.

[1]: https://github.com/tarantool/tarantool-python/issues/105#issuecomment-630907601

### DIFF
--- a/tarantool/schema.py
+++ b/tarantool/schema.py
@@ -10,32 +10,83 @@ from tarantool.utils import (
     integer_types,
 )
 from tarantool.error import (
+    Error,
     SchemaError,
     DatabaseError
 )
 import tarantool.const as const
 
 
+class RecursionError(Error):
+    """Report the situation when max recursion depth is reached.
+
+       This is internal error for <to_unicode_recursive> caller
+       and it should be re-raised properly be the caller.
+    """
+
+
+def to_unicode(s):
+    if isinstance(s, bytes):
+        return s.decode(encoding='utf-8')
+    return s
+
+
+def to_unicode_recursive(x, max_depth):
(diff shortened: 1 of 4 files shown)


Totktonada:
```python
assert max_depth > 0
```

@artembo suggested to don't use `assert` this way: it may be disabled in production use. I agreed and found that SchemaError should be raised here (it is internal error in fact and we'll define it properly in the error hierarhy within #174). I added RecursionError for technical matters: different `to_unicode_recursive()` calls should give different error messages when the recursion reaches its limit.

PR #173, opened 2020-08-26
Support msgpack 1.0.0

### LINKED ISSUE #155: New python-msgpack version brokes connector

Totktonada:
Fixes #155.
Supersedes PR #156.
Supersedes PR #167.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,15 @@ cache: false
 env:
     matrix:
       - TARGET=test
+        PYTHON_MSGPACK=msgpack-python==0.4.0
+      - TARGET=test
+        PYTHON_MSGPACK=msgpack==0.5.0
+      - TARGET=test
+        PYTHON_MSGPACK=msgpack==0.6.0
+      - TARGET=test
+        PYTHON_MSGPACK=msgpack==0.6.2
+      - TARGET=test
+        PYTHON_MSGPACK=msgpack==1.0.0
       - OS=el DIST=6
       - OS=el DIST=7
       - OS=fedora DIST=28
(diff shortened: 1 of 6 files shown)


tarantool/schema.py
```diff
@@ -54,14 +56,24 @@ def __init__(self, space_row, schema):
         self.arity = space_row[1]
         self.name = space_row[2]
         if isinstance(self.name, bytes):
-            self.name = self.name.decode()
+            self.name = self.name.decode(encoding='utf-8')
```
artembo, line 59: nit: Is it reasonable to set global variable for keeping the encoding value? You have eight the same part of the code.

Totktonada: It is about PR #172, but you're right. I'll create necessary wrappers.

Totktonada: I rewrote this part, see PR #172.

PR #181, opened 2020-11-25
Make code compatible with Python3

ligurio:


Totktonada:
See CI on the master branch: python 3 is supported and verified. So I don't understand a problem you tries to solve.

Say, `unittest.TestCase` contains both `assertRaisesRegexp()` and `assertRaisesRegex()` on Python 3.7.6. Even if it would not, we don't need a library to make the simple alias. `x.items()` from Python 2 is the same as `list(x.items())` in Python 3, but it does not meant that you need to follow `py2to3` everywhere it suggest you to do so. Mostly it is just doesn't necessary and a view on a mapping object is what you need.

Please, pay attention to don't break Python 2 code.

If I'm too early here, I suggest to mark the PR as draft.

ligurio:
> See CI on the master branch: python 3 is supported and verified. So I don't understand a problem you tries to solve.

Yep, code is actually already compatible with Python 3. There are only two places required to fix: print calls and exec call. Please review these two patches. CI is green.

> Please, pay attention to don't break Python 2 code.

Sure.

doc/conf.py
```diff
@@ -21,7 +21,7 @@
 # Read package version without importing it
 for line in open(os.path.join(os.path.dirname(os.path.abspath('.')), "tarantool", "__init__.py")):
     if line.startswith("__version__"):
-        exec line
+        exec(line)
```
Totktonada, line 24: A few words how do you meet a problem and how it looks would be useful. I would ever say that it is mandatory for this kind of changes. A problem may look obvious, but it does not. My initial guess was that the documentation building works only on sphinx that is run under Python 2 and it appears to be wrong. It worth to spend a minute to hold a problem and its appearance.

Surprisingly `python setup.py build_sphinx` works even for Python 3, but your change eliminates the following warning:

> WARNING: Support for evaluating Python 2 syntax is deprecated and will be removed in Sphinx 4.0. Convert /home/alex/projects/tarantool-meta/tarantool-python/doc/conf.py to Python 3 syntax.

So ironic: the tool is more backward compatible than the language it is written on.

BTW, I would add the 'doc' or the 'build' prefix (any of them looks applicable).

test/cluster-py/multi.test.py
```diff
@@ -18,11 +18,11 @@
 def check_connection(con):
     try:
         s = con.space('test')
-        print s.select()
+        print(s.select())
```
Totktonada, line 21: To be honest, I would glad to see a motivation here too (in the commit message). I guess I understand it: test-run is going to support Python 3 and it may be convenient to run tarantool-python's test using Python 3. But will I remember it after a year?..

I'll not insist, but, please, consider my humble opinion: those things are important, because they saves tons of time in a future.

doc/conf.py
ligurio: I couldn't reproduce a problem with warning:
```
(venv) [s.bronnikov@tarantool-core-dev-mcs1 tarantool-python]$ python --version
Python 3.6.8
(venv) [s.bronnikov@tarantool-core-dev-mcs1 tarantool-python]$ python setup.py build_sphinx
running build_sphinx
Running Sphinx v3.3.1
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
looking for now-outdated files... none found
no targets are out of date.
build succeeded.

The HTML pages are in build/sphinx/html.
```
Before introducing a change I read official Python documentatio. `exec` was a statement in Python 2.x [1] and it has been converted to a function in Python 3.x [2], change introduced in version 3.0 [3]. So I just followed to modern Python style.

BTW warning happens when Sphinx failed to exec python code. It tries to perform conversion 2to3 and print a warning, see [4].

1. https://docs.python.org/2/reference/simple_stmts.html#exec
2. https://docs.python.org/3/library/functions.html#exec
3. https://docs.python.org/3/whatsnew/3.0.html
4. https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/util/pycompat.py#L68-L88

ligurio: And I think it would be good idea to include building Sphinx documentation in CI. Motivation the same as for tests and other actions in CI - check that documentation build passed. What do you think?

test/cluster-py/multi.test.py
ligurio: Sure, added explanation to commit message and place here for your convenience:

```
commit cb33676fb9cf830c918ee61e09b532fa4aff60a1
Author: Sergey Bronnikov <sergeyb@tarantool.org>
Date:   Wed Nov 25 11:21:20 2020 +0300

    test: make print calls compatible with Python 3.x
    
    In a Python 3 'print' becomes a function, see [1].
    Patch makes print calls compatible with Python 3.
    
    1. https://docs.python.org/3/whatsnew/3.0.html#print-is-a-function
```

doc/conf.py
Totktonada: So the motivation is purely to keep in track with Python 3 versions. I generally discourage changes, which do not directed by a problem, because of fear of subtle behaviour changes. But this one looks quite straightforward, so okay for me.

Regarding docs in CI: I completely okay. In the past I propose / implement CI targets of this kind in other projects:

https://github.com/tarantool/tarantool-java/issues/244
https://github.com/tarantool/graphql.0/pull/273

test/cluster-py/multi.test.py
Totktonada: I would still guess that you had some problem around running on Python 3, but don't described it, if would read your commit message without any context. However, okay, I don't want to tire you with that.

Totktonada (approved):
LGTM.

Issue #183, opened 2020-12-24
PyPI release


a1tus:
Hello.

Are you planning to make a release with new features since 0.6.6?

In our case, we have to use commit id coz of breaking changes in msgpack 1.0.0.

Totktonada:
Sure.

Totktonada:
tarantool 0.7.1 was released.

Totktonada:
That was my mistake: I released [0.7.0](https://github.com/tarantool/tarantool-python/releases/tag/0.7.0) first, but then I found that it depends on `msgpack-python` (the old library name). So, I yanked the release on PyPI (AFAIU pip will not install it if this version is not requested explicitly) and released [0.7.1](https://github.com/tarantool/tarantool-python/releases/tag/0.7.1) with the correct dependency. Sorry for possible inconvenience.

a1tus:
@Totktonada great, thank you.

PR #185, opened 2020-12-28
pypi: correct msgpack library name, bump 0.7.1

Totktonada:
In short, the msgpack library changes its name and now offered as
'msgpack' rather than 'msgpack-python'. All new versions of the library
are shipped under the new name, so we should change our dependency
information.

See commit 148d12ea5dc4e999107211227a4da5f5a67bf62c ('travis-ci: verify
different msgpack library versions') for details.

Bump version to 0.7.1.

Follows up #155
Follows up #183

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,12 @@
+tarantool-python (0.7.1-0) unstable; urgency=medium
+
+    ## Overview
+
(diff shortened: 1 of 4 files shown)


Totktonada:
Nothing to verify — cancelled the CI.

PR #186, opened 2021-01-11
python3: make json.dumps compatible with Python 2

ligurio:
Python 3.4: Changed in version 3.4: Use (',', ': ') as default if
indent is not None. This behaviour broke a test box-py/call.test.py.

1. https://docs.python.org/3/library/json.html#json.dump


Totktonada (approved):
LGTM.

PR #187, opened 2021-01-18
python3: make conversion to string deterministic

ligurio:
Response object is a unpacked msgpack structure, returned by Tarantool.
Nor msgpack-python [1] that used in tarantool-python nor msgpack [2]
itself cannot guarantee an order of keys in unpacked dictionaries.
Therefore we have different keys order with running tests under Python 2
and Python 3, for example box-py/call.test.py. To workaround a problem
proposed a conversion of dictionaries in a tuple to lists before
printing Response object.

1. https://github.com/msgpack/msgpack-python/pull/164
2. https://github.com/msgpack/msgpack/issues/215

### DIFF
--- a/tarantool/response.py
+++ b/tarantool/response.py
@@ -29,6 +29,25 @@ from tarantool.error import (
     tnt_strerror
 )
 
+def deep_convert_dict(obj):
+    if isinstance(obj, dict):
+        ret = []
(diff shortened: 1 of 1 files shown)


ligurio:
Discussed and decided to workaround a problem in a test itself.
See https://github.com/tarantool/tarantool/issues/5538

PR #192, opened 2021-11-03
github-ci: add reusable testing workflow

ylobankov:
The idea of this workflow is to be a part of the 'integration.yml'
workflow from the tarantool repo to verify the integration of the
tarantool-python connector with an arbitrary tarantool version.

This workflow is not triggered on a push to the repo and cannot be run
manually since it has only the 'workflow_call' trigger. This workflow
will be included in the tarantool development cycle and called by the
'integration.yml' workflow from the tarantool project on a push to the
master and release branches for verifying integration of tarantool with
tarantool-python.

Part of tarantool/tarantool#6584
Part of tarantool/tarantool#5265
Part of tarantool/tarantool#6056

Related to tarantool/tarantool#6590

### DIFF
--- a/.github/workflows/reusable_testing.yml
+++ b/.github/workflows/reusable_testing.yml
@@ -1,0 +1,42 @@
+name: reusable_testing
+
+on:
+  workflow_call:
+    inputs:
+      artifact_name:
+        description: The name of the tarantool build artifact
+        default: ubuntu-focal
+        required: false
+        type: string
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone the tarantool-python connector
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.repository_owner }}/tarantool-python
+
+      - name: Download the tarantool build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+
+      - name: Install tarantool
+        # Now we're lucky: all dependencies are already installed. Check package
+        # dependencies when migrating to other OS version.
+        run: sudo dpkg -i tarantool*.deb
+
+      - name: Setup python3 for tests
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Install connector requirements
+        run: pip install -r requirements.txt
+
+      - name: Install test requirements
+        run: pip install -r requirements-test.txt
+
+      - run: make test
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  run_tests:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Clone the tarantool-python connector
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.repository_owner }}/tarantool-python
+          ref: ylobankov/add-reusable-testing-workflow
```
ylobankov, line 20: Will remove after testing is done.

ylobankov: Testing is done, removed.

Makefile
```diff
@@ -17,3 +17,5 @@ docs:
 	python setup.py build_sphinx
 docs-upload: docs
 	python setup.py upload_sphinx
+deps:
+	pip install pyyaml dbapi-compliance==1.15.0
```
Totktonada, line 21: I would prefer to add them into `requirements-test.txt` (it seems it is quite usual way). So we'll able to use it as on Linux as well as on Windows (in `appveyor.yml`) and will not need to remember to update it in two places.

.github/workflows/reusable_testing.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Clone the tarantool-python connector
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.repository_owner }}/tarantool-python
+
+      - name: Download the tarantool build artifact
+        uses: actions/download-artifact@v2
+        with:
+          name: ${{ inputs.artifact_name }}
+
```
Totktonada, line 25: Shouldn't we follow the recommendation to use the `setup-python` action?

> Using the `setup-python` action is the recommended way of using Python with GitHub Actions because it ensures consistent behavior across different runners and different versions of Python.

https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#specifying-a-python-version

Makefile
ylobankov: Ok, that makes sense.

.github/workflows/reusable_testing.yml
ylobankov: On the one hand, `ubuntu-20.04` GitHub runner already has Python3 inside and it works fine. So no setup actions are needed. On the other hand, I don't see anything bad in this recommendation, moreover, I see a few prons of it:
* Python version is pinned and we explicitly see it
* if we move to self-hosted runners, we will not be worry about Python

Makefile
Totktonada: Please, update it in the `appveyor.yml` file too.

ylobankov: Done.

Totktonada (approved):
I don't see any flaws.

PR #213, opened 2022-03-17
Introduce GitHub CI

### LINKED ISSUE #182: Setup GitHub Actions
Instead of Travis CI (which is going to be paid service) and AppVeyor (because GitHub Actions has Windows runners).

DifferentialOrange:
### Setup GitHub Actions testing pipeline

Before this patch, tarantool/tarantool-python CI was based on Travis CI (Linux runs) and Appveyor (Windows runs). This PR introduces GitHub Actions workflow files for both Linux and Windows test runs. Pipelines run for different supported tarantool, Python and msgpack package versions to ensure compatibility. tarantool instance is started in each Windows pipeline under WSL to run tests instead of using an external server (like in Appveyor). Since we start a new tarantool instance for each run, `clean()` procedure was removed. (The main reason to remove it was failing with `ER_DROP_FUNCTION: Can't drop function 1:
function is SQL built-in` error on newer 2.x on bootstrap.)

Testing pipeline for tarantool artifacts was also updated.

Travis CI and Appveyor badges were replaced with GitHub Actions badge.

### Skip flaky ping test on Windows

Ping test started to fail (in ~70% cases) after Windows CI migration from Appveyor to GitHub Actions. As a part of this migration, the test tarantool instance was changed from an instance started on a remote Linux server to an instance started on the same Windows server under WSL. Thus, request time has shortened and it supposedly caused the ping test to fail.

Python [documentation](https://docs.python.org/3/library/time.html#time.time) declares that

> ...though the time is always returned as a floating point number,
> not all systems provide time with a better precision than 1 second...

Particularly, this[ StackOverflow answer](https://stackoverflow.com/a/1938096/11646599) argues that

> For Linux and Mac precision is +- 1 microsecond or 0.001 milliseconds.
> Python on Windows uses +- 16 milliseconds precision due to clock
> implementation problems due to process interrupts.

based on Windows [documentation](https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/high-resolution-timers) [3].

Assuming that ping requests between the same server services can easilybe under 1 ms, it caused the test to fail.

This patch adds skip for the flaky test, refer to #214 for further development.

Closes #182, part of #214

After merging this PR, Appveyor workflows must be removed since they are not useful anymore.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,167 @@
+name: testing
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  run_tests_linux:
+    # We want to run on external PRs, but not on our own internal
+    # PRs as they'll be run by the push to the branch.
+    #
+    # The main trick is described here:
+    # https://github.com/Dart-Code/Dart-Code/pull/2375
+    if: github.event_name == 'push' ||
+      github.event.pull_request.head.repo.full_name != github.repository
+
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        tarantool:
+          - '1.10'
+          - '2.8'
+          - '2.x-latest'
+        python:
+          - '2.7'
+          - '3.5'
+          - '3.6'
+          - '3.7'
+          - '3.8'
+          - '3.9'
+          - '3.10'
+        msgpack-deps:
+          # latest msgpack will be installed as a part of requirements.txt
+          - ''
+
+        # Adding too many elements to three-dimentional matrix results in
+        # too many test cases. It causes GitHub webpages to fail with 
+        # "This page is taking too long to load." error. Thus we use
+        # pairwise testing.
+        include:
+          - tarantool: '2.8'
+            python: '3.10'
+            msgpack-deps: 'msgpack-python==0.4.0'
+          - tarantool: '2.8'
+            python: '3.10'
+            msgpack-deps: 'msgpack==0.5.0'
+          - tarantool: '2.8'
+            python: '3.10'
+            msgpack-deps: 'msgpack==0.6.2'
+          - tarantool: '2.8'
+            python: '3.10'
+            msgpack-deps: 'msgpack==1.0.0'
+
+    steps:
+      - name: Clone the connector
+        uses: actions/checkout@v2
+
+      - name: Install tarantool ${{ matrix.tarantool }}
+        if: matrix.tarantool != '2.x-latest'
+        uses: tarantool/setup-tarantool@v1
+        with:
+          tarantool-version: ${{ matrix.tarantool }}
+
+      - name: Install latest tarantool 2.x
+        if: matrix.tarantool == '2.x-latest'
+        run: |
+          curl -L https://tarantool.io/pre-release/2/installer.sh | sudo bash
+          sudo apt install -y tarantool tarantool-dev
+
+      - name: Setup Python for tests
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+
+      - name: Install specific version of msgpack package
+        if: startsWith(matrix.msgpack-deps, 'msgpack==') == true
+        run: |
+          pip install ${{ matrix.msgpack-deps }}
+
+      - name: Install specific version of msgpack-python package
+        # msgpack package is a replacement for deprecated msgpack-python.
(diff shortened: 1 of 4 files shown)


DifferentialOrange:
image

I can't see failed CI logs. I have filed a ticket to GitHub support, but there is no answer yet.

DifferentialOrange:
> image
> 
> I can't see failed CI logs. I have filed a ticket to GitHub support, but there is no answer yet.

It seems that having more than ~100 pipelines is too much of a load for GitHub website.

test/suites/lib/tarantool_python_ci.lua
```diff
@@ -343,5 +343,5 @@ end
 
 -- }}}
 
-clean()
+-- clean()
```
Totktonada, line 346: Let's remove the dead code.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,210 @@
...
+
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        tarantool:
+          - '1.10'
+          - '2.8'
+          - '2.9'
```
Totktonada, line 25: I would remove tarantool-2.9 (here and everywhere else), because it was never released.

.github/workflows/testing.yml
```diff
@@ -0,0 +1,210 @@
...
+          - '2.x-latest'
+        python:
+          - '2.7'
+          - '3.4'
+          - '3.5'
+          - '3.6'
+          - '3.7'
+          - '3.8'
+          - '3.9'
+          - '3.10'
```
Totktonada, line 124: The workflows on Windows are not cheap. Maybe keep [tarantool-1.10, tarantool-latest] × [python-2.7, python-3.10] plus different msgpack versions?

.github/workflows/reusable_testing.yml
```diff
@@ -28,12 +48,22 @@ jobs:
...
+      - name: Install msgpack python package
+        if: matrix.msgpack-deps != ''
+        run: pip install ${{ matrix.msgpack-deps }}
 
       - name: Install connector requirements
+        # msgpack package is a replacement for deprecated msgpack-python.
+        # To test compatibility with msgpack-python we must ignore
+        # requirements.txt install, since it will install newest msgpack.
+        # Beware that if any new dependency will be added to requirements.txt,
+        # this step will be invalid.
```
Totktonada, line 65: Maybe `sed -i -e 's/^msgpack.*$/${{ matrix.msgpack-deps }}/' requirements.txt` would be more reliable solution?

Totktonada (approved):
LGTM.

.github/workflows/reusable_testing.yml
```diff
@@ -12,6 +12,26 @@ on:
 jobs:
   run_tests:
     runs-on: ubuntu-20.04
+
+    strategy:
```
ylobankov, line 16: This workflow is mostly for testing of tarantool rather than tarantool-python. I am not sure that there is sense to do testing on different python versions and msgpack library versions here. Such testing makes sense to test the connector itself (like you do in testing.yml). IMHO, for testing of integration of tarantool and the connector the current variant of workflow is enough.

test/suites/lib/tarantool_python_ci.lua
DifferentialOrange: Fixed clean to work with 2.x. I decided not to change current behavior of clean execution since tarantool_python_ci.lua may be used in external projects like in https://github.com/tarantool/tarantool/issues/6965

.github/workflows/testing.yml
DifferentialOrange: Removed

.github/workflows/testing.yml
DifferentialOrange: Decided to leave only [tarantool-1.10, tarantool-2.8] × [python-2.7, python-3.10] since tarantool-latest is not stable and different msgpack version testing on Windows may be a bit too much.

.github/workflows/reusable_testing.yml
DifferentialOrange: Thank you, that's a really nice approach

.github/workflows/reusable_testing.yml
DifferentialOrange: Restored reusable_testing

ylobankov (commented):
LGTM, thanks!

PR #215, opened 2022-03-22
readme: update version in header and fix formatting

ligurio:
We don't need to explicitly set a version of Tarantool in header because
it may confuse Tarantool users. Python connector supports Tarantool with
versions below and upper version 1.6.0. The section below describes more
details about compatibility of Tarantool and connector versions. The first patch
removes a version in the header.

The second patch fixes items in a list.


Totktonada (approved):
LGTM.

PR #225, opened 2022-08-17
test: SSL with GOST cipher

DifferentialOrange:


DifferentialOrange:
Based on @oleg-jukovec comment, it's just a matter of openssl setup.

Issue #191, opened 2021-06-02
Is there a way to add custom type by myself? For example, uuid


TvoroG:
Hey! Thanks for the work!

I cannot figure out the right way to add custom type. `msgpack-python` supports ext type by `ext_hook` and `default` arguments to `Packer` and `Unpacker`, but we cannot pass them through tarantool connection. Am I wrong?

Totktonada:
It seems, you are right. I would make it extensible, more or less in the spirit of PR #190. I'll consider this issue as the feature request.

Issue #67, opened 2016-02-09
New manual for 0.5+ version driver


bigbes:
Our quick-start.{ru/en}.rst and guide.{ru/en}.rst are obsolete.

Superseeds #49 and #48.

Totktonada:
Several points to consider:

* We should revisit both guides and API docs.
* The task requires some level of Python understanding, reading throught the code and tests and manual testing.
* The docs should look nice on [readthedocs](https://tarantool-python.readthedocs.io/en/latest/).

DifferentialOrange:
One more crucial point: application developer's journey is incomprehensible. What's the life cycle? How should I process requests in code and its errors? Even tests are not useful due to `assertSequenceEqual`

PR #226, opened 2022-08-30
Support decimal extended type

### LINKED ISSUE #203: Support decimal
Protocol documentation: https://github.com/tarantool/doc/issues/992.
Python documentation: https://docs.python.org/3/library/decimal.html.

What is interesting: I don't see a way to construct Python's decimal from a binary form without encoding into a string first. Maybe I missed something.

DifferentialOrange:
Tarantool supports decimal type since version 2.2.1 [1]. This patch introduced the support of Tarantool decimal type in msgpack decoders and encoders. The Tarantool decimal type is mapped to the native Python decimal.Decimal type.

Tarantool decimal numbers have 38 digits of precision, that is, the total number of digits before and after the decimal point can be 38 [2]. If there are more digits arter the decimal point, the precision is lost. If there are more digits before the decimal point, error is thrown. In fact, there is also an exceptional case: if decimal starts with `0.`, 38 digits after the decimal point are supported without the loss of precision. msgpack encoder checks if everything is alright. If number is not a valid Tarantool decimal, the error is raised. If precision will be lost on conversion, warning is issued.

Any Tarantool decimal could be converted to a Python decimal without the loss of precision. Python decimals have its own user alterable precision (defaulting to 28 places), but it's related only to arithmetic operations: we can allocate 38-placed decimal disregarding of what decimal module configuration is used [3].

1. https://github.com/tarantool/tarantool/issues/692
2. https://www.tarantool.io/ru/doc/latest/reference/reference_lua/decimal/
3. https://docs.python.org/3/library/decimal.html

Closed #203

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ sophia
 # Vim Swap files
 *.sw[a-z]
 .idea
+
+venv/*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Decimal type support (#203).
 
 ### Changed
 - Bump msgpack requirement to 1.0.4 (PR #223).
--- a/tarantool/error.py
+++ b/tarantool/error.py
@@ -109,10 +109,20 @@ class ConfigurationError(Error):
     Error of initialization with a user-provided configuration.
     '''
 
+class MsgpackError(Error):
+    '''
+    Error with encoding or decoding of MP_EXT types
+    '''
+
+class MsgpackWarning(UserWarning):
+    '''
+    Warning with encoding or decoding of MP_EXT types
+    '''
 
 __all__ = ("Warning", "Error", "InterfaceError", "DatabaseError", "DataError",
            "OperationalError", "IntegrityError", "InternalError",
-           "ProgrammingError", "NotSupportedError")
+           "ProgrammingError", "NotSupportedError", "MsgpackError",
+           "MsgpackWarning")
 
 # Monkey patch os.strerror for win32
 if sys.platform == "win32":
--- a/tarantool/msgpack_ext/decimal.py
+++ b/tarantool/msgpack_ext/decimal.py
@@ -1,0 +1,228 @@
+from decimal import Decimal
+
+from tarantool.error import MsgpackError, MsgpackWarning, warn
+
+# https://www.tarantool.io/en/doc/latest/dev_guide/internals/msgpack_extensions/#the-decimal-type
+#
+# The decimal MessagePack representation looks like this:
+# +--------+-------------------+------------+===============+
+# | MP_EXT | length (optional) | MP_DECIMAL | PackedDecimal |
+# +--------+-------------------+------------+===============+
+#
+# PackedDecimal has the following structure:
+#
+#  <--- length bytes -->
+# +-------+=============+
+# | scale |     BCD     |
+# +-------+=============+
+#
+# Here scale is either MP_INT or MP_UINT.
+# scale = number of digits after the decimal point
+#
+# BCD is a sequence of bytes representing decimal digits of the encoded number
+# (each byte has two decimal digits each encoded using 4-bit nibbles), so
+# byte >> 4 is the first digit  and byte & 0x0f is the second digit. The
+# leftmost digit in the array is the most significant. The rightmost digit in
+# the array is the least significant.
+#
+# The first byte of the BCD array contains the first digit of the number,
+# represented as follows:
+#
+# |  4 bits           |  4 bits           |
+#    = 0x                = the 1st digit
+#
+# (The first nibble contains 0 if the decimal number has an even number of
+# digits.) The last byte of the BCD array contains the last digit of the number
+# and the final nibble, represented as follows:
+#
+# |  4 bits           |  4 bits           |
+#    = the last digit    = nibble
+#
+# The final nibble represents the number’s sign:
+#
+#     0x0a, 0x0c, 0x0e, 0x0f stand for plus,
+#     0x0b and 0x0d stand for minus.
+
+EXT_ID = 1
+
+TARANTOOL_DECIMAL_MAX_DIGITS = 38
+
+def get_mp_sign(sign):
+    if sign == '+':
+        return 0x0c
+
+    if sign == '-':
+        return 0x0d
+
+    raise RuntimeError
+
+def add_mp_digit(digit, bytes_reverted, digit_count):
+    if digit_count % 2 == 0:
+        bytes_reverted[-1] = bytes_reverted[-1] | (digit << 4)
+    else:
+        bytes_reverted.append(digit)
+
+def check_valid_tarantool_decimal(str_repr, scale, first_digit_ind):
+# Decimal numbers have 38 digits of precision, that is, the total number of
+# digits before and after the decimal point can be 38. If there are more
+# digits arter the decimal point, the precision is lost. If there are more
+# digits before the decimal point, error is thrown.
+#
+# Tarantool 2.10.1-0-g482d91c66
+#
+# tarantool> decimal.new('10000000000000000000000000000000000000')
+# ---
+# - 10000000000000000000000000000000000000
+# ...
+#
+# tarantool> decimal.new('100000000000000000000000000000000000000')
+# ---
+# - error: '[string "return VERSION"]:1: variable ''VERSION'' is not declared'
+# ...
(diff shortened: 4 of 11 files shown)


tarantool/ext_types/decimal.py
```diff
@@ -0,0 +1,226 @@
...
+        warn('Decimal encoded with loss of precision: ' + \
+             'Tarantool decimal supports a maximum of 38 digits.',
+             MsgpackWarning)
+        return False
+
+    return True
+
+def strip_decimal_str(str_repr, scale, first_digit_ind):
+    assert scale > 0
+    # Strip extra bytes
```
oleg-jukovec, line 131: No dot at the end of the comment.

tarantool/ext_types/decimal.py
```diff
@@ -0,0 +1,226 @@
...
+    str_repr = str_repr[:TARANTOOL_DECIMAL_MAX_DIGITS + 1 + first_digit_ind]
+
+    str_repr.rstrip('0')
+    str_repr.rstrip('.')
+    # Do not strips zeroes before the decimal point.
```
oleg-jukovec, line 136: Dot at the end of the comment.

tarantool/ext_types/decimal.py
```diff
@@ -0,0 +1,226 @@
...
+    sign = get_str_sign(data[-1] & 0x0f)
+
+    # Parse from tail since scale is counted from the tail.
+    digits_reverted = []
+
+    add_str_digit((data[-1] & 0xf0) >> 4, digits_reverted, scale)
+
+    for i in range(len(data) - 2, 0, -1):
+        add_str_digit(data[i] & 0x0f, digits_reverted, scale)
+        add_str_digit((data[i] & 0xf0) >> 4, digits_reverted, scale)
```
oleg-jukovec, line 216: It's a matter of taste, but it looks simply:
```suggestion
        add_str_digit((data[i]) >> 4, digits_reverted, scale)
```

tarantool/ext_types/decimal.py
```diff
@@ -0,0 +1,226 @@
...
+
+def decode(data):
+    scale = data[0] 
+
+    sign = get_str_sign(data[-1] & 0x0f)
+
+    # Parse from tail since scale is counted from the tail.
+    digits_reverted = []
+
+    add_str_digit((data[-1] & 0xf0) >> 4, digits_reverted, scale)
```
oleg-jukovec, line 212: It's up to you too.
```suggestion
    add_str_digit((data[-1]) >> 4, digits_reverted, scale)
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Decimal type support (#203).
```
oleg-jukovec, line 10: It would be great to add a section about supported types in the documentation.

.gitignore
```diff
@@ -17,3 +17,5 @@ sophia
 # Vim Swap files
 *.sw[a-z]
 .idea
+
+venv/*
```
oleg-jukovec, line 21: Up to you:

`dev: ignore venv folder` -> `git: ignore venv folder`

tarantool/ext_types/packer.py
```diff
@@ -0,0 +1,9 @@
+from decimal import Decimal
+from msgpack import ExtType
+
+import tarantool.ext_types.decimal as ext_decimal
```
oleg-jukovec, line 4: `ext_types` is a bit confusing. I would rename the directory. Up to you.
```suggestion
import tarantool.mgpack.decimal as ext_decimal
```
or
```suggestion
import tarantool.mgpack.ext.decimal as ext_decimal
```

oleg-jukovec (commented):
Thank you for the patch! A few notes:

tarantool/ext_types/packer.py
```diff
@@ -0,0 +1,9 @@
...
+
+def default(obj):
+    if isinstance(obj, Decimal):
+        return ExtType(ext_decimal.EXT_ID, ext_decimal.encode(obj))
+    raise TypeError("Unknown type: %r" % (obj,))
```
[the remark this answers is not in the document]
DifferentialOrange: Yeah, there is a separate issue for it: https://github.com/tarantool/tarantool-python/issues/191

Maybe it will be solved in #228 since this is where some controversy may raise. Actually it's not a big deal -- we just need to provide a couple of function argument to Connection object with out extended types packer/unpacker as a default.

tarantool/ext_types/decimal.py
DifferentialOrange: Thank you. It wasn't a matter of readability -- I just haven't worked with raw bytes for too long and missed out that it is excessive.

DifferentialOrange: Fixed

tarantool/ext_types/decimal.py
DifferentialOrange: Removed all dots since comments in the remaining code don't have one

tarantool/ext_types/decimal.py
DifferentialOrange: Removed all dots since comments in the remaining code don't have one

.gitignore
DifferentialOrange: Fair, changed

CHANGELOG.md
DifferentialOrange: It's rather simple and you don't have to do anything: if you receive a response with `decimal` from Tarantool, it would be converted to a Python `decimal` and vice versa. You don't need to do any additional imports like in go-tarantool.

On the other hand, you are right about the lack of documentation and examples. We will solve this as a part of https://github.com/tarantool/tarantool-python/issues/67 ticket which is planned after adding a datetime type (work in progress).

tarantool/ext_types/packer.py
DifferentialOrange: Any other msgpack-related code is scattered across the code so having a separate `msgpack` folder is redundant.

I rename the folder to `msgpack_ext_types`.

CHANGELOG.md
oleg-jukovec: > if you receive a response with `decimal` from Tarantool, it would be converted to a Python `decimal` and vice versa

This is where it may not be obvious. For example, I would expect some internal type in the package.

test/suites/test_ext_types.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def setUp(self):
+        # prevent a remote tarantool from clean our session
+        if self.srv.is_started():
+            self.srv.touch_lock()
+
+        self.adm("box.space['test']:truncate()")
+
+
+    valid_decimal_cases = [
```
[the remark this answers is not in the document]
DifferentialOrange: Thank you for your advice! I have added unit test. Actually, after adding them I have found a couple of bugs and fixed them.

tarantool/ext_types/packer.py
oleg-jukovec: `msgpack_ext` seems enough. Up to you.

test/suites/test_msgpack_ext_types.py
```diff
@@ -0,0 +1,425 @@
...
+            'python': decimal.Decimal('-1234567891234567890.0987654321987654321'),
+            'msgpack': (b'\x13\x01\x23\x45\x67\x89\x12\x34\x56\x78\x90' +
+                        b'\x09\x87\x65\x43\x21\x98\x76\x54\x32\x1d'),
+            'tarantool': "decimal.new('-1234567891234567890.0987654321987654321')",
+        },
+    ]
+
+    def test_decimal_msgpack_decode(self):
+        for i in range(len(self.valid_decimal_cases)):
+            with self.subTest(msg=str(i)):
```
oleg-jukovec, line 218: It would be easier in the future to find a failed case by value, rather than count a number:
```suggestion
            with self.subTest(msg=str(self.valid_decimal_cases[i]['python'])):
```
The same in all tests.

tarantool/ext_types/packer.py
DifferentialOrange: Used `msgpack_ext`, thank you

test/suites/test_msgpack_ext_types.py
DifferentialOrange: Used named test cases

CHANGELOG.md
DifferentialOrange: Type mapping and restrictions would be added in https://github.com/tarantool/tarantool-python/issues/67 solution

oleg-jukovec (approved):
LGTM

PR #227, opened 2022-08-31
Support UUID extended type

### LINKED ISSUE #202: Support UUID
Protocol documentation: https://github.com/tarantool/doc/issues/1227.

DifferentialOrange:
Merge after #226 

Tarantool supports UUID type since version 2.4.1 [1]. This patch introduced the support of Tarantool UUID type in msgpack decoders and encoders. The Tarantool UUID type is mapped to the native Python uuid.UUID type.

1. https://github.com/tarantool/tarantool/issues/4268

Closed #202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Decimal type support (#203).
+- UUID type support (#202).
 
 ### Changed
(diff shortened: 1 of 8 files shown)


oleg-jukovec (approved):
Thank you for the patch! LGTM, but I would split test_msgpack_ext.py into two files. It may become too big in the future.

DifferentialOrange:
> but I would split test_msgpack_ext.py into two files.

Reworked

PR #230, opened 2022-09-09
Support datetime interval extended type and arithmetic

### LINKED ISSUE #229: Support interval type
Follows up #204

See https://www.tarantool.io/ru/doc/latest/dev_guide/internals/msgpack_extensions/#the-interval-type

DifferentialOrange:
### msgpack: support datetime interval extended type

Tarantool supports datetime interval type since version 2.10.0 [1]. This patch introduced the support of Tarantool interval type in msgpack decoders and encoders.

Tarantool datetime interval objects are decoded to `tarantool.Interval` type. `tarantool.Interval` may be encoded to Tarantool interval objects.

You can create `tarantool.Interval` objects either from msgpack data or by using the same API as in Tarantool:
```python
di = tarantool.Interval(year=-1, month=2, day=3,
                        hour=4, minute=-5, sec=6,
                        nsec=308543321,
                        adjust=tarantool.IntervalAdjust.NONE)
```

Its attributes (same as in init API) are exposed, so you can use them if needed.

datetime, numpy and pandas tools doesn't seem to be sufficient to cover all adjust cases supported by Tarantool.

This patch does not yet introduce the support of datetime interval arithmetic.

1. https://github.com/tarantool/tarantool/issues/5941

Part of #229

### msgpack: support datetime interval arithmetic

Support datetime and interval arithmetic with the same rules as in Tarantool [1].

Valid operations:
  - `tarantool.Datetime` + `tarantool.Interval` = `tarantool.Datetime`
  - `tarantool.Datetime` - `tarantool.Interval` = `tarantool.Datetime`
  - `tarantool.Datetime` - `tarantool.Datetime` = `tarantool.Interval`
  - `tarantool.Interval` + `tarantool.Interval` = `tarantool.Interval`
  - `tarantool.Interval` - `tarantool.Interval` = `tarantool.Interval`

Since `tarantool.Interval` could contain `month` and `year` fields and such operations could be ambiguous, you can use `adjust` field to tune the logic.

- `tarantool.IntervalAdjust.NONE` -- only truncation toward the end of month performed (default mode).

  ```
  >>> dt = tarantool.Datetime(year=2022, month=3, day=31)
  datetime: Timestamp('2022-03-31 00:00:00'), tz: ""
  >>> di = tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.NONE)
  >>> dt + di
  datetime: Timestamp('2022-04-30 00:00:00'), tz: ""
  ```

- `tarantool.IntervalAdjust.EXCESS` -- overflow mode, without any snap or truncation to the end of month, straight addition of days in month, stopping over month boundaries if there is less number of days.

  ```
  >>> dt = tarantool.Datetime(year=2022, month=1, day=31)
  datetime: Timestamp('2022-01-31 00:00:00'), tz: ""
  >>> di = tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.EXCESS)
  >>> dt + di
  datetime: Timestamp('2022-03-02 00:00:00'), tz: ""
  ```

- `tarantool.IntervalAdjust.LAST` -- mode when day snaps to the end of month, if happens.

  ```
  >>> dt = tarantool.Datetime(year=2022, month=2, day=28)
  datetime: Timestamp('2022-02-28 00:00:00'), tz: ""
  >>> di = tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.LAST)
  >>> dt + di
  datetime: Timestamp('2022-03-31 00:00:00'), tz: ""
  ```

Tarantool does not yet correctly support subtraction of datetime objects with different timezones [2] and addition of intervals to datetimes with non-fixed offset timezones [3]. tarantool-python implementation support them, but it could be reworked later if core team choose another solution.

1. https://github.com/tarantool/tarantool/wiki/Datetime-Internals#interval-arithmetic
2. https://github.com/tarantool/tarantool/issues/7698
3. https://github.com/tarantool/tarantool/issues/7700

Closes #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   You may use `tz` property to get timezone name of a datetime object.
 
+- Datetime interval type support and tarantool.Interval type (#229).
+
+  Tarantool datetime interval objects are decoded to `tarantool.Interval`
+  type. `tarantool.Interval` may be encoded to Tarantool interval
+  objects.
+
+  You can create `tarantool.Interval` objects either from msgpack
+  data or by using the same API as in Tarantool:
+
+  ```python
+  di = tarantool.Interval(year=-1, month=2, day=3,
+                          hour=4, minute=-5, sec=6,
+                          nsec=308543321,
+                          adjust=tarantool.IntervalAdjust.NONE)
+  ```
+
+  Its attributes (same as in init API) are exposed, so you can
+  use them if needed.
+
 ### Changed
 - Bump msgpack requirement to 1.0.4 (PR #223).
   The only reason of this bump is various vulnerability fixes,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,54 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Its attributes (same as in init API) are exposed, so you can
   use them if needed.
 
+- Datetime interval arithmetic support (#229).
+
+  Valid operations:
+  - `tarantool.Datetime` + `tarantool.Interval` = `tarantool.Datetime`
+  - `tarantool.Datetime` + `tarantool.Interval` = `tarantool.Datetime`
+  - `tarantool.Datetime` - `tarantool.Datetime` = `tarantool.Interval`
+  - `tarantool.Interval` + `tarantool.Interval` = `tarantool.Interval`
+  - `tarantool.Interval` - `tarantool.Interval` = `tarantool.Interval`
+
+  Since `tarantool.Interval` could contain `month` and `year` fields
+  and such operations could be ambiguous, you can use `adjust` field
+  to tune the logic. The behavior is the same as in Tarantool, see
+  [Interval arithmetic RFC](https://github.com/tarantool/tarantool/wiki/Datetime-Internals#interval-arithmetic).
+
+  - `tarantool.IntervalAdjust.NONE` -- only truncation toward the end of
+    month performed (default mode).
+
+    ```python
+    >>> dt = tarantool.Datetime(year=2022, month=3, day=31)
+    datetime: Timestamp('2022-03-31 00:00:00'), tz: ""
+    >>> di = tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.NONE)
+    >>> dt + di
+    datetime: Timestamp('2022-04-30 00:00:00'), tz: ""
+    ```
+
+  - `tarantool.IntervalAdjust.EXCESS` -- overflow mode, without any snap
+    or truncation to the end of month, straight addition of days in month,
+    stopping over month boundaries if there is less number of days.
+
+    ```python
+    >>> dt = tarantool.Datetime(year=2022, month=1, day=31)
+    datetime: Timestamp('2022-01-31 00:00:00'), tz: ""
+    >>> di = tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.EXCESS)
+    >>> dt + di
+    datetime: Timestamp('2022-03-02 00:00:00'), tz: ""
+    ```
+
+  - `tarantool.IntervalAdjust.LAST` -- mode when day snaps to the end of month,
+    if happens.
+
+    ```python
+    >>> dt = tarantool.Datetime(year=2022, month=2, day=28)
+    datetime: Timestamp('2022-02-28 00:00:00'), tz: ""
+    >>> di = tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.LAST)
+    >>> dt + di
+    datetime: Timestamp('2022-03-31 00:00:00'), tz: ""
+    ```
+
 ### Changed
 - Bump msgpack requirement to 1.0.4 (PR #223).
   The only reason of this bump is various vulnerability fixes,
--- a/tarantool/msgpack_ext/types/datetime.py
+++ b/tarantool/msgpack_ext/types/datetime.py
@@ -6,6 +6,8 @@ import pytz
 import tarantool.msgpack_ext.types.timezones as tt_timezones
 from tarantool.error import MsgpackError
 
+from tarantool.msgpack_ext.types.interval import Interval, Adjust
+
 # https://www.tarantool.io/en/doc/latest/dev_guide/internals/msgpack_extensions/#the-datetime-type
 #
 # The datetime MessagePack representation looks like this:
@@ -47,6 +49,7 @@ BYTEORDER = 'little'
 NSEC_IN_SEC = 1000000000
 NSEC_IN_MKSEC = 1000
 SEC_IN_MIN = 60
+MONTH_IN_YEAR = 12
 
 def get_bytes_as_int(data, cursor, size):
     part = data[cursor:cursor + size]
@@ -167,6 +170,83 @@ class Datetime():
         else:
             self._datetime = datetime
             self._tz = ''
+
+    def _interval_operation(self, other, sign=1):
+        self_dt = self._datetime
+
+        # https://github.com/tarantool/tarantool/wiki/Datetime-Internals#date-adjustions-and-leap-years
+        months = other.year * MONTH_IN_YEAR + other.month
+
+        res = self_dt + pandas.DateOffset(months = sign * months)
+
+        # pandas.DateOffset works exactly like Adjust.NONE
+        if other.adjust == Adjust.EXCESS:
+            if self_dt.day > res.day:
+                res = res + pandas.DateOffset(days = self_dt.day - res.day)
+        elif other.adjust == Adjust.LAST:
+            if self_dt.is_month_end:
+                # day replaces days
+                res = res.replace(day = res.days_in_month)
+
+        res = res + pandas.Timedelta(weeks = sign * other.week,
+                                     days = sign * other.day,
+                                     hours = sign * other.hour,
+                                     minutes = sign * other.minute,
+                                     seconds = sign * other.sec,
+                                     microseconds = sign * (other.nsec // NSEC_IN_MKSEC),
+                                     nanoseconds = sign * (other.nsec % NSEC_IN_MKSEC))
+
+        if res.tzinfo is not None:
+            tzoffset = compute_offset(res)
+        else:
+            tzoffset = 0
+        return Datetime(year=res.year, month=res.month, day=res.day,
+                        hour=res.hour, minute=res.minute, sec=res.second,
+                        nsec=res.nanosecond + res.microsecond * NSEC_IN_MKSEC,
+                        tzoffset=tzoffset, tz=self.tz)
+
+    def __add__(self, other):
+        if not isinstance(other, Interval):
+            raise TypeError(f"unsupported operand type(s) for +: '{type(self)}' and '{type(other)}'")
+
+        return self._interval_operation(other, sign=1)
+
+    def __sub__(self, other):
+        if isinstance(other, Datetime):
+            self_dt = self._datetime
+            other_dt = other._datetime
+
+            # Tarantool datetime subtraction ignores timezone info, but it is a bug:
+            #
+            # Tarantool 2.10.1-0-g482d91c66
+            #
+            # tarantool> datetime.new{tz='MSK'} - datetime.new{tz='UTC'}
+            # ---
+            # - +0 seconds
+            # ...
+            #
+            # Refer to https://github.com/tarantool/tarantool/issues/7698
+            # for possible updates.
+
+            if self_dt.tzinfo != other_dt.tzinfo:
+                other_dt = other_dt.tz_convert(self_dt.tzinfo)
+
+            self_nsec = self_dt.microsecond * NSEC_IN_MKSEC + self_dt.nanosecond
+            other_nsec = other_dt.microsecond * NSEC_IN_MKSEC + other_dt.nanosecond
+
+            return Interval(
+                year = self_dt.year - other_dt.year,
+                month = self_dt.month - other_dt.month,
+                day = self_dt.day - other_dt.day,
+                hour = self_dt.hour - other_dt.hour,
+                minute = self_dt.minute - other_dt.minute,
+                sec = self_dt.second - other_dt.second,
+                nsec = self_nsec - other_nsec,
+            )
+        elif isinstance(other, Interval):
+            return self._interval_operation(other, sign=-1)
+        else:
+            raise TypeError(f"unsupported operand type(s) for -: '{type(self)}' and '{type(other)}'")
 
     def __eq__(self, other):
         if isinstance(other, Datetime):
--- a/tarantool/msgpack_ext/types/interval.py
+++ b/tarantool/msgpack_ext/types/interval.py
@@ -106,6 +106,70 @@ class Interval():
             self.nsec = nsec
             self.adjust = adjust
 
+    def __add__(self, other):
+        if not isinstance(other, Interval):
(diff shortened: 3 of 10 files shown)


tarantool/msgpack_ext/types/datetime.py
```diff
@@ -168,6 +171,78 @@ def __init__(self, data=None, *, timestamp=None, year=None, month=None,
...
+        else:
+            tzoffset = 0
+        return Datetime(year=res.year, month=res.month, day=res.day,
+                        hour=res.hour, minute=res.minute, sec=res.second,
+                        nsec=res.nanosecond + res.microsecond * NSEC_IN_MKSEC,
+                        tzoffset=tzoffset, tz=self.tz)
+
+    def __sub__(self, other):
+        if not isinstance(other, Datetime):
+            raise TypeError(f"unsupported operand type(s) for -: '{type(self)}' and '{type(other)}'")
```
oleg-jukovec, line 213: Tarantool supports subtraction of intervals from a Datetime. We could do the same.

tarantool/msgpack_ext/types/interval.py
```diff
@@ -0,0 +1,213 @@
...
+            value = getattr(self, field_name)
+
+            if field_name == 'adjust':
+                value = value.value
+
+            if value != 0:
+                buf = buf + msgpack.packb(field_id) + msgpack.packb(value)
+                count = count + 1
+
+        buf = msgpack.packb(count) + buf
```
oleg-jukovec, line 211: Up to you, but the class violates Single Responsibility Principle: it mixes encoding/decoding and the interval-logic code.

test/suites/test_datetime_arithmetic.py
```diff
@@ -0,0 +1,317 @@
...
+
+        self.assertEqual(case['arg_1'] - case['arg_2'], case['res'])
+
+    @skip_or_run_datetime_test
+    @unittest.expectedFailure # See https://github.com/tarantool/tarantool/issues/7698
+    def test_tarantool_datetime_sub_different_timezones(self):
+        case = self.datetime_sub_different_timezones_case
+
+        self.assertSequenceEqual(self.con.call('sub', case['arg_1'], case['arg_2']),
+                                         [case['res']])
```
oleg-jukovec, line 215: Wil the testsl fail on a new versions of Tarantool with fixes? It seems better to skip the test.

test/suites/test_datetime_arithmetic.py
```diff
@@ -0,0 +1,317 @@
...
+        'month_non_last_day_last_adjust': {
+            'arg_1': tarantool.Datetime(year=2009, month=3, day=30),
+            'arg_2': tarantool.Interval(month=2, adjust=tarantool.IntervalAdjust.LAST),
+            'res': tarantool.Datetime(year=2009, month=5, day=30),
+        },
+        'month_overflow_last_day_last_adjust': {
+            'arg_1': tarantool.Datetime(year=2009, month=2, day=28),
+            'arg_2': tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.LAST),
+            'res': tarantool.Datetime(year=2009, month=3, day=31),
+        },
```
oleg-jukovec, line 274: Here you could add more cases:

1. month <= 0
2. NONE + src_days <= dst_days
3. EXCESS + src_days <= dst_days
4. LAST + src_days > dst_days

tarantool/msgpack_ext/types/interval.py
```diff
@@ -0,0 +1,213 @@
...
+        # If built not from msgpack data, set argument values later. 
+        self.year = 0
+        self.month = 0
+        self.week = 0
+        self.day = 0
+        self.hour = 0
+        self.minute = 0
+        self.sec = 0
+        self.nsec = 0
+        self.adjust = Adjust(0)
```
oleg-jukovec, line 67: Is it ok for python to access fields directly without getters/setters/propery?

test/suites/test_datetime_arithmetic.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def setUp(self):
+        # prevent a remote tarantool from clean our session
+        if self.srv.is_started():
+            self.srv.touch_lock()
+
+        self.adm("box.space['test']:truncate()")
+
+
+    interval_cases = {
```
oleg-jukovec, line 64: It looks like tests for the Interval class, not for the Datetime.

test/suites/test_datetime_arithmetic.py
```diff
@@ -0,0 +1,317 @@
...
+        'res': tarantool.Datetime(year=2008, month=7, day=1, tz='Europe/Moscow'),
+    }
+
+    def test_python_datetime_add_winter_time_switch(self):
+        case = self.datetime_add_winter_time_switch_case
+
+        self.assertEqual(case['arg_1'] + case['arg_2'], case['res'])
+
+    @skip_or_run_datetime_test
+    @unittest.expectedFailure # See https://github.com/tarantool/tarantool/issues/7700
```
oleg-jukovec, line 305: Wil the testsl fail on a new versions of Tarantool with fixes? It seems better to skip the test.

test/suites/test_datetime_arithmetic.py
```diff
@@ -0,0 +1,317 @@
...
+            with self.subTest(msg=name):
+                case = self.datetime_add_cases[name]
+
+                self.assertSequenceEqual(self.con.call('add', case['arg_1'], case['arg_2']),
+                                         [case['res']])
+
+    datetime_add_winter_time_switch_case = {
+        'arg_1': tarantool.Datetime(year=2008, month=1, day=1, tz='Europe/Moscow'),
+        'arg_2': tarantool.Interval(month=6),
+        'res': tarantool.Datetime(year=2008, month=7, day=1, tz='Europe/Moscow'),
```
oleg-jukovec, line 296: It will be more obvious with a hour values.

test/suites/test_interval.py
```diff
@@ -0,0 +1,206 @@
...
+import pytz
+
+from tarantool.msgpack_ext.packer import default as packer_default
+from tarantool.msgpack_ext.unpacker import ext_hook as unpacker_ext_hook
+
+from .lib.tarantool_server import TarantoolServer
+from .lib.skip import skip_or_run_datetime_test
+from tarantool.error import MsgpackError
+
+class TestSuite_Interval(unittest.TestCase):
```
oleg-jukovec, line 20: We don't test Add/Sub in the file, just encoding/decoding.

test/suites/test_interval.py
```diff
@@ -0,0 +1,206 @@
...
+
+        self.adm("box.space['test']:truncate()")
+
+
+    cases = {
+        'year': {
+            'python': tarantool.Interval(year=1),
+            'msgpack': (b'\x02\x00\x01\x08\x01'),
+            'tarantool': r"datetime.interval.new({year=1})",
+        },
```
oleg-jukovec, line 59: There is no test cases for:

tarantool.Interval(data=foobar)

oleg-jukovec (commented):
Thank you for the patch! I don't see a critical problems, just few issues.

CHANGELOG.md
```diff
@@ -67,6 +67,71 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  - `tarantool.IntervalAdjust.LAST` -- mode when day snaps to the end of month,
+    if happens.
+
+    ```python
+    >>> dt = tarantool.Datetime(year=2022, month=2, day=28)
+    datetime: Timestamp('2022-02-28 00:00:00'), tz: ""
+    >>> di = tarantool.Interval(month=1, adjust=tarantool.IntervalAdjust.LAST)
+    >>> dt + di
+    datetime: Timestamp('2022-03-31 00:00:00'), tz: ""
+    ```
```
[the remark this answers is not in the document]
DifferentialOrange: I think it would be useful to have an explicit documentation about basic behavior.

tarantool/msgpack_ext/types/datetime.py
DifferentialOrange: Thank you. I had completely forgot about it. Will fix it up

test/suites/test_datetime_arithmetic.py
DifferentialOrange: Yes, it will. I think it's better to wait until the bug would be fixed and then replace xfail with conditional xfail based on version rather than set a skip and forget to ever run the test.

test/suites/test_datetime_arithmetic.py
DifferentialOrange: By "Datetime arithmetic" and I implied both datetime and interval interactions. I think it is alright since in Tarantool the classes are built with `datetime.new` and `datetime.interval` new, but if you have some ideas about what name is better suited to describe both of them -- I would be glad to hear it.

test/suites/test_datetime_arithmetic.py
DifferentialOrange: https://github.com/tarantool/tarantool-python/pull/230#discussion_r983429755

test/suites/test_datetime_arithmetic.py
oleg-jukovec: Maybe we should run the test only on CI? (by a environment variable or an another way that unittest support)

On the one hand, we'll see when our tests break. On the other hand, a user will not be confused with a failed test after updating Tarantool.

CHANGELOG.md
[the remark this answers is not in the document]
oleg-jukovec: I understand why it's here. But we can just refer to the Tarantool implementation, since we don't provide any differences from the implementation.

It may help to avoid "broken phone". Up to you.

test/suites/test_datetime_arithmetic.py
oleg-jukovec: I think it will be more obvious to leave two files: `test_datetime.py` for the Datetime class API and `test_interval.py` for the Interval class API. As example I have three files:

1. test_datetime.py
2. test_interval.py
3. test_datetime_arithmetic.py

Where should I look test for `+` operation of the Interval class?

DifferentialOrange: Well, now all `+` tests is in `test_datetime_arithmetic.py`.

With your approach, it would be `Interval + Interval` in `test_interval.py` and `Datetime + Interval` in... `test_interval.py`? `test_datetime.py`? Separate file `test_datetime_arithmetic.py`? To be honest, I don't think there could be non-confusing approaches.

test/suites/test_datetime_arithmetic.py
DifferentialOrange: > NONE + src_days <= dst_days
> EXCESS + src_days <= dst_days
> LAST + src_days > dst_days

Adjustment is not affect days addition since there is only one way to add days. Or am I misunderstood something?

CHANGELOG.md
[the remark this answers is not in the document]
DifferentialOrange: I get what your intention is, but I think the convenience of a user is more important. Moreover, we have our own Tarantool-like implementation of adjustments here, so if anything would change and "phone" will "broke", we will need to update not only our documentation but our implementation as well.

test/suites/test_datetime_arithmetic.py
oleg-jukovec: test_datetime.py: `+` is the method of the Datetime class.

CHANGELOG.md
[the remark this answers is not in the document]
oleg-jukovec: > if anything would change and "phone" will "broke", we would need to update not only our documentation but our implementation as well.

I think we should update the implementation in anyway.

test/suites/test_datetime_arithmetic.py
oleg-jukovec: You know it. I know it. But the code does not know it. We can to test the behavior in the unit-test. It doesn't overcomplicate anything.

tarantool/msgpack_ext/types/datetime.py
DifferentialOrange: Supported `datetime` - `interval`

tarantool/msgpack_ext/types/interval.py
DifferentialOrange: I think we leave it as is for now since the design is the same as in `Datetime`

test/suites/test_datetime_arithmetic.py
DifferentialOrange: > On the other hand, a user will not be confused with a failed test after updating Tarantool.

I think that when one should get this fail, it would be the moment we need to replace xfail with conditional xfail based on version.

test/suites/test_datetime_arithmetic.py
DifferentialOrange: month <= 0 now covered as positive month subtraction

test/suites/test_datetime_arithmetic.py
DifferentialOrange: Added

test/suites/test_interval.py
DifferentialOrange: Seems like a part of https://github.com/tarantool/tarantool-python/pull/230#discussion_r983314441 discussion

test/suites/test_interval.py
DifferentialOrange: Added

CHANGELOG.md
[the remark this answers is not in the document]
DifferentialOrange: I've added RFC link, but I still don't want to remove adjustment examples from the CHANGELOG.

test/suites/test_datetime_arithmetic.py
oleg-jukovec: Add a task to a sprint, fix it, and then release a new version... It may take a long time for users to get a working version.

It's not a good idea to make CI easier for us in the moment, but make life harder for users in the future.

test/suites/test_datetime_arithmetic.py
DifferentialOrange: Well, the solution was more obvious than I expected.

Moved all interval methods tests to `test_interval.py` and datetime methods tests to `test_datetime.py`

test/suites/test_interval.py
DifferentialOrange: Unknown interval field and re-raising Adjust ValueError already tested in `test_unknown_field_decode` and `test_unknown_adjust_decode`. I don't think it's worth to test them one more time but with direct `data=bytes''` call.

DifferentialOrange: Regarding the first one, I've added `test_Interval_bytes_init_ignore_other_fields`

oleg-jukovec: However, there is a difference between the unit test for the constructor and for the decode function from msgpack.

oleg-jukovec (approved):
LGTM.

test/suites/test_datetime_arithmetic.py
DifferentialOrange: I don't get how choosing between `xfail` and `skip` affects user with release versions.

DifferentialOrange: If you are running module tests, you are a developer. As a developer, you may either ignore a couple of failing tests after you find out what's the reason (it shouldn't be complicated since relevant issue link is provided), or report the issue to us (to be honest, I'm sure that developer would already be one of us), or fix it (it's an hour long work together with all required stuff like commit messages and PR).

DifferentialOrange: I would be glad to provide `skip` tests with conditions, but these conditions not yet exist. Make a test that always will skip until one remembers to "uncomment it" is also doesn't seems like a good approach to me.

test/suites/test_interval.py
DifferentialOrange: Well, ok, I've added these tests + test on `data` is `bytes`

PR #233, opened 2022-09-26
UNIX socket test

DifferentialOrange:
Add test for `tarantool.connect` with UNIX socket. Add TarantoolServer option to start a Tarantool instance which listens to UNIX socket.

### DIFF
--- a/test/suites/lib/tarantool_server.py
+++ b/test/suites/lib/tarantool_server.py
@@ -122,7 +122,8 @@ class TarantoolServer(object):
                 ssl_key_file=None,
                 ssl_cert_file=None,
                 ssl_ca_file=None,
-                ssl_ciphers=None):
+                ssl_ciphers=None,
+                create_unix_socket=False):
         if os.name == 'nt':
             from .remote_tarantool_server import RemoteTarantoolServer
             return RemoteTarantoolServer()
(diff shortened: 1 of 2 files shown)


test/suites/lib/tarantool_server.py
```diff
@@ -36,6 +36,26 @@ def find_port(port=None):
...
+    sock.bind(bind_addr)
+    sock.close()
+
+    return True
+
+def find_socket(bind_num=None):
+    if bind_num is None:
+        bind_num = random.randrange(0, 99)
+    while bind_num < 99:
+        bind_addr = f'/tmp/tarantool_{bind_num}.sock'
```
oleg-jukovec, line 53: At first glance it will be more portable to use a per session `tmpdir()` from pytest instead of `/tmp/.

https://code-maven.com/pytest-fixtures-temporary-directory
https://github.com/tarantool/tt/blob/master/test/conftest.py

oleg-jukovec (approved):
Thank you for the patch!

test/suites/lib/tarantool_server.py
DifferentialOrange: Thank you, reworked with `tempfile.NamedTemporaryFile`

oleg-jukovec (approved):
Thank you for the patch!

Issue #238, opened 2022-10-07
Publish documentation with CD


DifferentialOrange:
Follows #237

DifferentialOrange:
Now webhooks and applications are set up. Docs should build for each master update. They are also built for pull requests so it is possible to review doc without any local build.

PR #241, opened 2022-10-10
Fix module installation

DifferentialOrange:
Part of #238 (required for correct autoversion display)

Refer to commits for detailed info.

This PR fixed installation commands like
```bash
pip install git+https://github.com/tarantool/tarantool-python.git@DifferentialOrange/gh-238-fix-install
```

Before this patch, `setup.py` missed several requirements and `msgpack_ext` folder was missing from the package.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -214,11 +214,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        # Use reduced test matrix cause Windows pipelines are long.
         tarantool:
           - '1.10'
           - '2.8'
           - '2.10.0.g0a5ce0b9c-1'
         python:
+          - '3.6'
           - '3.10'
 
     steps:
@@ -292,9 +294,11 @@ jobs:
       fail-fast: false
 
       matrix:
+        # Use reduced test matrix cause Windows pipelines are long.
         tarantool:
           - '2.10.0.g0a5ce0b9c-1'
         python:
+          - '3.6'
           - '3.10'
     steps:
       - name: Clone the connector repo
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -148,6 +148,57 @@ jobs:
         env:
           TEST_TNT_SSL: ${{ matrix.tarantool == '2.10.0-1-gfa775b383-r486-linux-x86_64' }}
 
+  run_tests_with_install_linux:
+    # We want to run on external PRs, but not on our own internal
+    # PRs as they'll be run by the push to the branch.
+    #
+    # The main trick is described here:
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -50,6 +50,14 @@ def read(*parts):
...
         return fp.read()
 
+def get_dependencies(file):
+    root = os.path.dirname(os.path.realpath(__file__))
+    requirements = os.path.join(root, file)
+    result = []
+    if os.path.isfile(requirements):
+        with open(requirements) as f:
+            result = f.read().splitlines()
+    return result
```
oleg-jukovec, line 60: It could be easier to understand that something is going wrong:
```suggestion
def get_dependencies(file):
    root = os.path.dirname(os.path.realpath(__file__))
    requirements = os.path.join(root, file)
    if os.path.isfile(requirements):
        with open(requirements) as f:
            return f.read().splitlines()
    raise RuntimeError("Unable to get dependencies from file " + file)
```

setup.py
```diff
@@ -64,6 +72,7 @@ def find_version(*file_paths):
     name="tarantool",
     packages=["tarantool"],
     package_dir={"tarantool": os.path.join("tarantool")},
```
oleg-jukovec, line 74: https://docs.python.org/3/distutils/setupscript.html#listing-whole-packages
```suggestion
    package_dir={"tarantool": "tarantool"},
```

oleg-jukovec (commented):
We could add test install to CI.

setup.py
DifferentialOrange: Thanks, added

setup.py
DifferentialOrange: Thanks, tried to use it

DifferentialOrange:
>We could add test install to CI.

Added in separate commit

.github/workflows/testing.yml
```diff
@@ -224,3 +275,80 @@ jobs:
...
+    runs-on: windows-2022
+
+    strategy:
+      fail-fast: false
+
+      matrix:
+        tarantool:
+          - '2.10.0.g0a5ce0b9c-1'
+        python:
+          - '3.10'
```
oleg-jukovec, line 298: Are there any restrictions on the python version on Windows? Does it need to be documented?

DifferentialOrange: Well, there are no restrictions, it's more like "test only minimal required" since Windows pipelines are rather long.

oleg-jukovec: I think then you need to add 3.6 (3.6, 3.10) since we claim it as a minimal supported version.

DifferentialOrange: That's a good comment, added

PR #243, opened 2022-10-13
Support iproto feature discovery

### LINKED ISSUE #206: Support IPROTO_ID (feature discovery)
Protocol: https://github.com/tarantool/doc/issues/2419.
net.box API (as the reference API): https://github.com/tarantool/doc/issues/2418.

There is also https://github.com/tarantool/doc/issues/2416, but I would leave extended errors out of Q1 scope: I would like to revisit errors in tarantool and at least have good vision around this area at whole.

DifferentialOrange:
### code-health: update request type constant name

Modern Tarantool documentation uses IPROTO_REQUEST_TYPE name for request type instead of IPROTO_CODE [1].

1. https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/

Part of #206

### code-health: use hexadecimal codes for requests

Modern Tarantool documentation uses hexadecimal codes for request codes [1].

1. https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/

Part of #206

### code-health: expose DatabaseError properties

Provide explicit handles to extract DatabaseError exceptions properties. Before this patch, one could extract Tarantool error code only through internal `args` property.

Unfortunately, current exceptions API is inconsistent. Code uses `DatabaseError(code, message)`, `NetworkError(message)` and
`NetworkError(exc)`, while NetworkError is a child class of DatabaseError.

This patch shouldn't break current behavior.

Part of #206

### iproto: support feature discovery

Since version 2.10.0 Tarantool supports feature discovery [1]. Client can send the schema version and supported features and receive server-side schema version and supported features information to tune its behavior.

After this patch, the request would be send on `connect`. Connector will use protocol version that is minimal of connector version (now it's 4) and server version. Feature would be enabled if both client and server supports it (for now client does not support any features from the list). Unknown request type error response is expected for pre-2.10.0 versions. In this case, protocol version would be `None` and no features would be enabled.

1. https://github.com/tarantool/tarantool/issues/6253

Closes #206


GRISHNOV (approved):
Hi! LGTM

PR #244, opened 2022-10-13
api: any msgpack supported type as a request key

### LINKED ISSUE #240: Extention types as keys
There is an explicit type checks for SELECT, UPDATE and DELETE operation keys: https://github.com/tarantool/tarantool-python/blob/5ffe902f1009a7c7418e2b6a1a504a81d9eb6f4c/tarantool/utils.py#L26

Extended types are not supported yet:

```python
>>> conn.select('space', uuid.uuid4())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/georgymoiseev/Development/github/tarantool/tarantool-python/tarantool/connection.py", line 1519, in select
    key = check_key(key, select=True)
  File "/home/georgymoiseev/Development/github/tarantool/tarantool-python/tarantool/utils.py", line 51, in check_key
    assert isinstance(key, supported_types)
AssertionError
```

DifferentialOrange:
Inserting a tuple with connector does not have any type restrictions on its contents: any MessagePack supported type is allowed. If there are any type violations on the Tarantool side, server would return the corresponding error. There is no reason to do any explicit type checks for request keys.

This patch fixed using extended types as keys. Tarantool 2.10 does not support indexing intervals, so there are no tests for INTERVAL extension type.

Closes #240

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,6 +146,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Package build (#238).
+- Allow any MessagePack supported type as a request key (#240).
 
 ## 0.9.0 - 2022-06-20
 
--- a/tarantool/connection.py
+++ b/tarantool/connection.py
@@ -86,9 +86,9 @@ from tarantool.error import (
 )
 from tarantool.schema import Schema
 from tarantool.utils import (
-    check_key,
     greeting_decode,
     version_id,
+    wrap_key,
     ENCODING_DEFAULT,
 )
 
@@ -1220,7 +1220,7 @@ class Connection(ConnectionInterface):
         .. _delete: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/delete/
         """
 
-        key = check_key(key)
+        key = wrap_key(key)
         if isinstance(space_name, str):
             space_name = self.schema.get_space(space_name).sid
         if isinstance(index, str):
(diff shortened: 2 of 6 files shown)


oleg-jukovec:
> If there are any type violations on the Tarantool side, server would return the corresponding error.

Is there a test for this behavior?

DifferentialOrange:
> > If there are any type violations on the Tarantool side, server would return the corresponding error.
> 
> Is there a test for this behavior?

To be honest, I don't think it's worth it because it would be a test for Tarantool only.

- Create some schema with eval
- Insert something
- Select something with invalid key type, get format mismatch

This behavior haven't even changed with this patch.

GRISHNOV (approved):
Hi! LGTM

DifferentialOrange:
> > > If there are any type violations on the Tarantool side, server would return the corresponding error.
> > 
> > 
> > Is there a test for this behavior?
> 
> To be honest, I don't think it's worth it because it would be a test for Tarantool only.
> 
>     * Create some schema with eval
> 
>     * Insert something
> 
>     * Select something with invalid key type, get format mismatch
> 
> 
> This behavior haven't even changed with this patch.

In other words, there is no such thing as "unsupported type of a request key". Anything could be a key until it matches space format (and connector does not do any excessive format validations between tuples/keys and spaces). There is such a thing as "unsupported type for a primary index", but it is not related to connector at all.

oleg-jukovec:
> In other words, there is no such thing as "unsupported type of a request key".

I get an idea with an unsupported MessagePack extension ID.

DifferentialOrange:
> > In other words, there is no such thing as "unsupported type of a request key".
> 
> I get an idea with an unsupported MessagePack extension ID. Tarantool should return the error in this case.

Yeah, it should be like this. But it is still not related to this patch. This patch is simply about removing unnecessary type checkup. I think it was here for Tarantool 1.5

Issue #196, opened 2021-11-18
Automatic master discovery


Totktonada:
* Poll `box.info.ro` on each instance every X seconds.
* Add `mode = 'any' / 'ro' / 'prefer_ro' / 'prefer_rw' / 'rw'` parameter for `call`, `eval`, `execute`, `select`, `ping`, `insert`, `replace`, `delete`, `update`, `upsert`.
* Fail fast: if we have no rw instance ATM -> report the fail to a caller.
* Fail fast: got `ER_READ_ONLY` -> report it to a caller.

RFC is [here](https://docs.google.com/document/d/1Yc6LlWmwGS2wJTE66pnqXH2VvVNf257J8McgCnw5EWA).

Related: https://github.com/tarantool/go-tarantool/issues/113.

PR #245, opened 2022-10-13
Support box.error

### LINKED ISSUE #232: Support error type
Since version [2.4.1](https://www.tarantool.io/en/doc/latest/release/2.4.1/), responses for errors have extra information following what was described in [Box protocol – responses for errors](https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/#box-protocol-responses-error). This is a “compatible” enhancement, because clients that expect old-style server responses should ignore map components that they do not recognize. Notice, however, that there has been a renaming of a constant: formerly IPROTO_ERROR in ./box/iproto_constants.h was 0x31, now IPROTO_ERROR is 0x52 and IPROTO_ERROR_24 is 0x31.

`MP_ERROR` type is 3.

See [doc](https://www.tarantool.io/en/doc/latest/dev_guide/internals/msgpack_extensions/#the-error-type) for more details.

DifferentialOrange:
### setup: clarify Python versions

We run tests with Python 3.6 and newer. Python 3.5 has reached its end of life two years ago [1]. Our code already doesn't work on Python 3.5 [2].

1. https://endoflife.date/python
2. https://github.com/tarantool/tarantool-python/actions/runs/3243224481/jobs/5317632662

Part of #232

### deps: backport dataclasses for Python 3.6

Dataclasses would be a really convenient approach to provide box.error support. They are supported in Python since 3.7, but this package make it able to introduce their support in Python 3.6 which we still support.

Part of #232

### api: ConnectionPool support for Python 3.6

After backporting dataclasses, it became possible to use ConnectionPool on Python 3.6.

Follows #196

### msgpack: improve unknown ext type error message

Part of #232

### iproto: support errors extra information

Since Tarantool 2.4.1, iproto error responses contain extra info with backtrace. After this patch, DatabaseError would contain `extra_info` property, if it was provided.

Error extra information is parsed based on common encoder/decoder rules. String fields are converted to either `str` or `bytes` based on `encoding` mode.

1. https://www.tarantool.io/en/doc/latest/dev_guide/internals/box_protocol/#responses-for-errors

Part of #232

### msgpack: support error extension type

Tarantool supports error extension type since version 2.4.1 [1], encoding was introduced in Tarantool 2.10.0 [2]. This patch introduces the support of Tarantool error extension type in msgpack decoders and encoders.

Tarantool error extension type objects are decoded to `tarantool.BoxError` type. `tarantool.BoxError` may be encoded to Tarantool error extension type objects.

Error extension type internals are the same as errors extra information: the only difference is that extra information is encoded as a separate error dictionary field and error extension type objects is encoded as MessagePack extension type objects.

Error extension type objects are parsed based on common encoder/decoder rules. String fields are converted to either `str` or `bytes` based on `encoding` mode.

The only way to receive an error extension type object from Tarantool is to receive an explicitly built `box.error` object: either from
`return box.error.new(...)` or a tuple with it. All errors raised within Tarantool (including those raised with `box.error(...)`) are encoded based on the same rules as simple errors due to backward compatibility.

It is possible to create error extension type objects with Python code, but it not likely to be really useful since most of their fields is
computed on error initialization on the server side (even for custom error types):

```python
tarantool.BoxError(
    type='ClientError',
    file='[string "                     local err = box.error.ne..."]',
    line=1,
    message='Unknown error',
    errno=0,
    errcode=0,
)
```

1. https://github.com/tarantool/tarantool/issues/4398
2. https://github.com/tarantool/tarantool/issues/6433

Closes #232


DifferentialOrange:
There is some minor CI issue after rebase, I think it is still possible to review current version.

tarantool/msgpack_ext/error.py
```diff
@@ -0,0 +1,52 @@
...
+
+    :param packer: msgpack packer to encode error dictionary.
+    :type packer: :class:`msgpack.Packer`
+
+    :return: Encoded error.
+    :rtype: :obj:`bytes`
+    """
+
+    err_map = encode_box_error(obj)
+    return packer.pack(err_map)
```
oleg-jukovec, line 34: It feels inconsistency in `encode` implementations for extension types:

error - a separate function
datetime, interval - an unnecessary method
uuid - a field (implementation itself?)
decimal - implementation itself

You could unify it here before a release.

tarantool/types.py
```diff
@@ -0,0 +1,141 @@
+"""
+Additional Tarantool type definitions.
+"""
+
+import typing
+from dataclasses import dataclass
+
+@dataclass
+class BoxError():
```
oleg-jukovec, line 9: Why not `tarantool/msgpack_ext/types/error.py` ?

DifferentialOrange: Because it's actually not a msgpack extension type. You can get `BoxError` object as additional info (type is dictionary, not MP_EXT N) in plain error, and I think it would be 99% of cases when one will work with `BoxError`. But you also can get MP_EXT 3 type with the same dictionary data.

DifferentialOrange:
> There is some minor CI issue after rebase, I think it is still possible to review current version.

Fixed in separate PR #251 (this PR could be reviewed independently).

tarantool/msgpack_ext/error.py
DifferentialOrange: > uuid - a field (implementation itself?)

It's "implementation itself": uuid just natively maps to byte sequence both with Tarantool and Python.

DifferentialOrange: > datetime, interval - an unnecessary method

Ok, let's make code more consistent. Since it would seem rather weird as a part of this PR, I have opened a separate one (which expected to be merged before this one): please, review #252

DifferentialOrange: Regarding
> error - a separate function
see https://github.com/tarantool/tarantool-python/pull/245#discussion_r1003072944 . This separate function does not create MP_EXT object. Moreover, there are cases when we need to decode only box.error dict without MP_EXT types (on the other way, there is no such cases for encode).

DifferentialOrange: #252 merged

PR #247, opened 2022-10-19
iproto: support feature push

### LINKED ISSUE #201: Consider IPROTO_PUSH support
See discussions in https://github.com/tarantool/go-tarantool/issues/67 and https://github.com/tarantool/tarantool/issues/6107.

GRISHNOV:
Adds support for receiving out-of-band messages from a server that uses box.session.push call.

Implemented for methods: `call`, `eval`, `select`, `insert`, `replace`, `update`, `upsert`, `delete`.
To work with out-of-band messages, 2 new optional arguments are used:
* `on_push` [function] - callback, launched with the received data for each out-of-band message. Two arguments for this callback are expected:
    +  the first is the received from an out-of-band message data.
    +  the second is `on_push_ctx`,  variable for recording the result of the callback work.
* `on_push_ctx` - result of the `on_push` work can be written to this variable.

Below is an example of the proposed API with method `call` and `insert`. In the described example, before the end of the `call` and `insert`, out-of-band messages are processed via specified callback.

For comparison: implementation in the [lua version](https://www.tarantool.io/ru/doc/latest/reference/reference_lua/box_session/push/).

<table>
<thead>
<tr>
<th>Client</th>
<th>Server</th>
</tr>
</thead>
<tbody>

<tr>
<td></td><td>
<pre><code class="language-lua">fiber = require('fiber')
box.cfg({listen = 3301})
box.schema.user.grant(
    'guest',
    'read,write,execute',
    'universe'
)
function server_function()
    x = {0,0}
    while x[1] < 3 do
        x[1] = x[1] + 1
        fiber.sleep(1)
        box.session.push(x)
    end
    fiber.sleep(1)
    return x
end</code></pre>
</td>
</tr>

<tr>
<td>
<pre><code class="language-python">import tarantool
<br>
def callback(data, on_push_ctx=[]):
    print('run callback with data: ', data)
    data[0][1] = data[0][1] + 1
    on_push_ctx.append(data)
<br>
callback_res = []
<br>
conn = tarantool.connect(port=3301)
res = conn.call(
    'server_function',
    on_push=callback, 
    on_push_ctx=callback_res
)
<code>
# receiving out-of-band messages, 
# the conn.call is not finished yet.</code>
<b>
>>> run callback with data:  [[1, 0]]
>>> run callback with data:  [[2, 0]]
>>> run callback with data:  [[3, 0]]
</b><code>
# the conn.call is finished now.</code>
<br>
print(res)
<b>>>> [3, 0]</b>
<br>
print(callback_res)
<b>>>>[[[1, 1]], [[2, 1]], [[3, 1]]]</b>
<br>

<tr>
<td></td><td>
<pre><code class="language-lua">box.schema.create_space(
 'tester', {
  format = {
    {name = 'id', type = 'unsigned'},
    {name = 'name', type = 'string'},
  }
})
box.space.tester:create_index(
 'primary_index', {
  parts = {
    {field = 1, type = 'unsigned'},
  }
})
function on_replace_callback()
    x = {0,0}
    while x[1] < 300 do
        x[1] = x[1] + 100
        box.session.push(x)
    end
    return x
end
box.space.tester:on_replace(
    on_replace_callback
)</code></pre>

</td>
</tr>

<tr>
<td>
<pre><code class="language-python">callback_res = []
<br>
conn_pool = tarantool.ConnectionPool(
        [{'host':'localhost', 'port':3301}],
        user='guest')
<br>
res = conn_pool.insert(
    'tester',
    (1, 'Mike'),
    mode=tarantool.Mode.PREFER_RO,
    on_push=callback,
    on_push_ctx=callback_res,
)
<code>
# receiving out-of-band messages, 
# the conn_pool.insert is not finished yet.</code>
<b>
>>> run callback with data:  [[100, 0]]
>>> run callback with data:  [[200, 0]]
>>> run callback with data:  [[300, 0]]
</b><code>
# the conn_pool.insert is finished now.</code>
<br>
print(res)
<b>>>> [1, 'Mike']</b>
<br>
print(callback_res)
<b>>>>[[[100, 1]], [[200, 1]], [[300, 1]]]</b>
<br>

</tbody>
</table>

Closes #201


Totktonada:
In the Lua API `on_push` is a callback (called at receiving of a pushed value) and `on_push_ctx` is a value to pass to the callback (context). As I see from the PR description, here the API is different: pushed values are just stored in a provided collection. One can't instantly react on a push using a code during a long request. Is it intentional?

GRISHNOV:
> In the Lua API `on_push` is a callback (called at receiving of a pushed value) and `on_push_ctx` is a value to pass to the callback (context). As I see from the PR description, here the API is different: pushed values are just stored in a provided collection. One can't instantly react on a push using a code during a long request. Is it intentional?

I have updated the solution and description in PR. Now it is possible to use callback, which will be called streaming when out-of-band messages are received before the end of the main call (for example, the end of `call` or `eval`).

tarantool/connection.py
```diff
@@ -767,12 +769,30 @@ def _send_request_wo_reconnect(self, request):
...
-                self._socket.sendall(bytes(request))
+                if not iproto_chunk_detected:
+                    # If the last received message is out-of-band, 
+                    # the request will not be sent.
+                    self._socket.sendall(bytes(request))
                 response = request.response_class(self, self._read_response())
-                break
+                if response._code == IPROTO_CHUNK:
+                    # Сase of receiving an out-of-band message.
+                    pushed_data.append(copy.deepcopy(response._data))
```
DifferentialOrange, line 785: As far as I understand, there is no pushed_data in Lua API. I'm not sure we should add it as well. If one would want to process it or collect it, he always may use `on_push` and `on_push_res`.

DifferentialOrange: Oh, there is no `on_push_res` in Lua as well. I think it's better to use exactly the same api with `on_push` and `on_push_ctx` with `None` default values for both.

This way, we won't add any extra load for request which wants to ignore possible push messages (except for `if callable()` check), yet we will support all possible cases (including "collect all messages to a list"). Then we just provide a couple of small examples in doc about how to work with iproto.

tarantool/connection.py
```diff
@@ -767,12 +769,30 @@ def _send_request_wo_reconnect(self, request):
...
+                    if callable(on_push):
+                        # Callback function with data from out-of-band 
+                        # message is being called.
+                        on_push_res.append(on_push(copy.deepcopy(response._data)))
```
DifferentialOrange, line 789: Since the response for `IPROTO_CHUNK` isn't exposed anywhere else, I think it will be ok to remove copy if we get rid of `pushed_data` and `on_push` for a single `on_push(data, on_push_ctx)`.

tarantool/connection.py
```diff
@@ -1368,7 +1446,7 @@ def ping(self, notime=False):
             return "Success"
         return t1 - t0
 
-    def select(self, space_name, key=None, *, offset=0, limit=0xffffffff, index=0, iterator=None):
+    def select(self, space_name, key=None, *, offset=0, limit=0xffffffff, index=0, iterator=None, **kwargs):
```
DifferentialOrange, line 1449: Is there any examples of receiving push messages in select/replace/etc?

DifferentialOrange (commented):
Thank you for your draft!

I think we may implement push in a way even more similar to Lua one and describe basic cases with tests and small examples. Rework should be relatively small.

tarantool/connection.py
```diff
@@ -914,7 +934,7 @@ def flush_schema(self):
         self.schema.flush()
         self.load_schema()
 
-    def call(self, func_name, *args):
+    def call(self, func_name, *args, **kwargs):
```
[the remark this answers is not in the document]

tarantool/connection.py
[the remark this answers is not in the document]
GRISHNOV: Done. I also added to the PR description an example with using `tarantool.ConnectionPool` with implemented functionality

tarantool/connection.py
GRISHNOV: Done, also updated the PR description in this part

tarantool/connection.py

tarantool/connection.py
GRISHNOV: I added an example with `insert` via `tarantool.ConnectionPool` to the PR description. I plan to transfer the description from the current PR to `/docs/source/quick-start.rst`

GRISHNOV:
Thank you for your feedback! If the current version of the draft is suitable, I will start writing tests and documentation

tarantool/connection.py
```diff
@@ -767,12 +770,30 @@ def _send_request_wo_reconnect(self, request):
...
+        iproto_chunk_detected = False
+        
         response = None
         while True:
             try:
-                self._socket.sendall(bytes(request))
+                if not iproto_chunk_detected:
+                    # If the last received message is out-of-band, 
+                    # the request will not be sent.
+                    self._socket.sendall(bytes(request))
```
oleg-jukovec, line 782: It will be more clear:
```suggestion
                if not request_sended:
                    # If the last received message is out-of-band, 
                    # the request will not be sent.
                    self._socket.sendall(bytes(request))
                    request_sended = True
```
or even extract the send from the loop.

tarantool/connection.py
```diff
@@ -767,12 +770,30 @@ def _send_request_wo_reconnect(self, request):
...
                 response = request.response_class(self, self._read_response())
-                break
+                if response._code == IPROTO_CHUNK:
+                    # Сase of receiving an out-of-band message.
+                    if callable(on_push):
```
oleg-jukovec, line 786: If not callable?

GRISHNOV: Done, now if `on_push` is not None and is not callable, then `raise TypeError`

tarantool/connection.py
GRISHNOV: AFAIU, calling the `self._socket.sendall`  in this loop may be required several times, therefore extracting `self._socket.sendall` from loop is a problem

DifferentialOrange: +1, to be honest, `iproto_chunk_detected` is rather hard to catch on from the first glance.

DifferentialOrange:
> If the current version of the draft is suitable,

I think it is. To fix docs build, I recommend you to rebase on master branch. Moreover, I think it's better to rebase on yet unmerged https://github.com/tarantool/tarantool-python/pull/251 since it contains fixes for potential test fails.

doc/quick-start.rst
```diff
@@ -200,3 +200,151 @@ read-write and read-only pool instances:
...
+
+To work with out-of-band messages, 2 optional arguments are used in 
+the methods listed above:
+
+ * `on_push` - callback, launched with the received data for each out-of-band message. 
+Two arguments for this callback are expected:
+   
+   * the first is the received from an out-of-band message data.
+   * the second is `on_push_ctx`,  variable for recording the result of the callback work.
+ * `on_push_ctx` - result of the `on_push` work can be written to this variable.
```
oleg-jukovec, line 223: I would not be so categorical at the description. The callback can just use the value.

GRISHNOV: Done

tarantool/connection.py
GRISHNOV: To be honest, I don't see such opportunity

oleg-jukovec: Tarantool checks schema version on commit:

https://github.com/tarantool/tarantool/blob/master/src/box/iproto.cc#L2000

So it may cause the error after IPROTO_CHUNK. You could try to reproduce it with:

1. 1 terminal:

```Lua
box.schema.create_space(
 'tester', {
  format = {
    {name = 'id', type = 'unsigned'},
    {name = 'name', type = 'string'},
  }
})
box.space.tester:create_index(
 'primary_index', {
  parts = {
    {field = 1, type = 'unsigned'},
  }
})
function on_replace_callback()
    x = {0,0}
    while x[1] < 300 do
        x[1] = x[1] + 100
        box.session.push(x)
        fiber.sleep(100)
    end
    return x
end
box.space.tester:on_replace(
    on_replace_callback
)
```

2. 2 terminal:
```python
res = conn_pool.insert(
    'tester',
    (1, 'Mike'),
    mode=tarantool.Mode.PREFER_RO,
    on_push=callback,
    on_push_ctx=callback_res,
)
```

3. 1 terminal

Change a schema.

4. Wait ~100 seconds

docs/source/quick-start.rst
```diff
@@ -200,3 +200,151 @@ read-write and read-only pool instances:
...
+
+.. code-block:: python
+
+    callback_res = []
+
+    conn_pool = tarantool.ConnectionPool(
+            [{'host':'localhost', 'port':3301}],
+            user='guest')
+
+    res = conn_pool.insert(
```
oleg-jukovec, line 329: Connection.insert or ConnectionPool.insert?

GRISHNOV: Done

tarantool/connection.py
DifferentialOrange: Actually, what behavior did we expect in case of receiving a SchemaReloadException?

tarantool/connection_pool.py
```diff
@@ -750,6 +756,12 @@ def eval(self, expr, *args, mode=None):
         :param mode: Request mode.
         :type mode: :class:`~tarantool.Mode`
 
+        :param on_push: Refer to
+            :paramref:`~tarantool.Connection.call.params.on_push`.
```
DifferentialOrange, line 760:
```suggestion
            :paramref:`~tarantool.Connection.eval.params.on_push`.
```

and, please, check others too

tarantool/const.py
```diff
@@ -41,6 +41,7 @@
 
 IPROTO_GREETING_SIZE = 128
 IPROTO_BODY_MAX_LEN = 2147483648
+IPROTO_CHUNK=0x80
```
DifferentialOrange, line 44: const is a bit messy, but I think it should go after `IPROTO_FEATURES = 0x55` and with spaces around ` = `

test/suites/lib/skip.py
```diff
@@ -162,3 +162,11 @@ def skip_or_run_datetime_test(func):
 
     return skip_or_run_test_pcall_require(func, 'datetime',
                                       'does not support datetime type')
+
+def skip_or_run_push_test_via_conn_pool(func):
```
DifferentialOrange, line 166: Just use `skip_or_run_conn_pool_test`

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+    return srv
+
+
+# Callback for on_push arg (for testing purposes).
+def push_callback(data, on_push_ctx=[]):
+                data[0][1] = data[0][1] + 1
+                on_push_ctx.append(data)
+
+
+@unittest.skipIf(sys.platform.startswith("win"),
```
DifferentialOrange, line 83: Mesh=ConnectionMesh tests, and I don't think they are

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+                 'Mesh tests on windows platform are not supported')
+class TestSuite_Push(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(self):
+        print(' PUSH '.center(70, '='), file=sys.stderr)
+        print('-' * 70, file=sys.stderr)
+
+    def setUp(self):
```
DifferentialOrange, line 92: It is required to create a server for each test case?

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+                warnings.simplefilter('ignore', category=NetworkWarning)
+                resp = conn.call(
+                    'server_function',
+                    on_push=push_callback, 
+                    on_push_ctx=callback_res
+                )
+            self.assertEqual(callback_res, [[[1, 1]], [[2, 1]], [[3, 1]]])
+            self.assertEqual(resp.data[0], [3, 0])
+
+        conn = tarantool.connect(host=self.host, port=self.port, user='test', password='test')
```
DifferentialOrange, line 111: Why not open it in `setUp`?

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+        # Check push from instance.
+        assert_push_call(conn)
+        conn.close()
+        # Stop instance.
+        self.srv.stop()
```
DifferentialOrange, line 116: It definitely should be in `teardown`

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+
+        with self.assertRaises(NetworkError):
```
DifferentialOrange, line 118: What do we want to check with this one?

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+            with warnings.catch_warnings():
+                warnings.simplefilter('ignore', category=NetworkWarning)
+                conn.ping()
+
+    
+    def test_00_02_push_insert_via_connect(self):
```
DifferentialOrange, line 149: Actually, the reason why basic (`test_dml.py`) tests are enumerated is that they are codependent and cannot be reordered. So it is not like you should enumerate your suit as well. (But you can is you wish, we didn't have a unified style yet.)

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+        on_replace_callback
+    )
+    """)
+
+    return srv
+
+
+# Callback for on_push arg (for testing purposes).
+def push_callback(data, on_push_ctx=[]):
+                data[0][1] = data[0][1] + 1
```
DifferentialOrange, line 79: Indentation is a bit weird

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+        with self.assertRaises(NetworkError):
+            with warnings.catch_warnings():
+                warnings.simplefilter('ignore', category=NetworkWarning)
+                conn.ping()
+
+    
+    def test_00_02_push_insert_via_connect(self):
+        def assert_push_insert(conn):
+            callback_res = []
+            with warnings.catch_warnings():
```
DifferentialOrange, line 152: Why are you catching any warnings here?

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+                warnings.simplefilter('ignore', category=NetworkWarning)
```
DifferentialOrange, line 153: And why we should receive any `NetworkWarning`s here?

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+            self.assertEqual(resp.data[0], [3, 0])
+
+        conn_pool = tarantool.ConnectionPool([{'host':self.host, 'port':self.port}], user='test', password='test')
+        # Check push from instance.
+        assert_push_eval(conn_pool)
+        conn_pool.close()
+        # Stop instance.
+        self.srv.stop()
+
+    @skip_or_run_push_test_via_conn_pool
```
DifferentialOrange, line 329: It's like you can do a test matrix here (I couldn't say that there are good examples of it, but at least there is some working examples, like in decimal/datetime)

test/suites/test_push.py
DifferentialOrange: Oh, it is already there. That's one more reason why it shouldn't be there too.

DifferentialOrange (commented):
Thank you for your PR! I have left several comments. If there is any questions about test infrastructure (currently it's a mess) or doc (it should be in shape after last PRs), feel free to ask me.

docs/source/quick-start.rst
```diff
@@ -200,3 +200,151 @@ read-write and read-only pool instances:
...
+        'tester',
+        (1, 'Mike'),
+        mode=tarantool.Mode.PREFER_RO,
```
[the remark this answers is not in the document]
DifferentialOrange: `PREFER_RO` is a bit specific mode -- the only case when it's useful is when you want to read from replica so there won't be any excessive load for read-write instances which are busy with writing. I think it's better to use `ANY` for calls if there is no any reason to do otherwise and default mode where there is one.

docs/source/quick-start.rst
```diff
@@ -200,3 +200,151 @@ read-write and read-only pool instances:
...
+:meth:`~tarantool.Connection.insert`, :meth:`~tarantool.Connection.replace`, 
+:meth:`~tarantool.Connection.update`, :meth:`~tarantool.Connection.upsert`, 
+:meth:`~tarantool.Connection.delete`.
+
+To work with out-of-band messages, 2 optional arguments are used in 
+the methods listed above:
+
+ * `on_push` - callback, launched with the received data for each out-of-band message. Two arguments for this callback are expected:
+   
+    the first is the received from an out-of-band message data.
```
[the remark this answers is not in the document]
GRISHNOV: Done

docs/source/quick-start.rst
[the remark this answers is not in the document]
GRISHNOV: Done

tarantool/connection.py
```diff
@@ -64,6 +64,7 @@
     IPROTO_FEATURE_ERROR_EXTENSION,
     IPROTO_FEATURE_WATCHERS,
     IPROTO_FEATURE_GRACEFUL_SHUTDOWN,
+    IPROTO_CHUNK
```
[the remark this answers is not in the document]

tarantool/connection.py
```diff
@@ -766,14 +768,20 @@ def _read_response(self):
...
+    def _send_request_wo_reconnect(self, request, on_push=None, on_push_ctx=None):
         """
         Send request without trying to reconnect.
         Reload schema, if required.
 
         :param request: Request to send.
         :type request: :class:`~tarantool.request.Request`
 
+        :param on_push: Сallback for processing out-of-band messages.
+        :type on_push: :obj:`function`, optional
```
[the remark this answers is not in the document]
GRISHNOV: Done

tarantool/connection.py
```diff
@@ -766,14 +768,20 @@ def _read_response(self):
...
+
+        :param on_push_ctx: Сontext for working with on_push callback.
+        :type on_push_ctx: :obj:`Any`, optional
```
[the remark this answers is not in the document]

tarantool/connection_pool.py
GRISHNOV: Done

tarantool/const.py
GRISHNOV: Done

test/suites/lib/skip.py
GRISHNOV: Done

test/suites/test_push.py
GRISHNOV: Ouch, fixed

test/suites/test_push.py
GRISHNOV: I tried to do something like this:

```python
@skip_or_run_conn_pool_test
def setUp(self):
        # Open connection and connection pool to instance.
        self.conn = tarantool.connect(host=self.host, port=self.port, 
                                      user='test', password='test')
        self.conn_pool = tarantool.ConnectionPool([{'host':self.host, 'port':self.port}], 
                                                     user='test', password='test')
```
But then I didn't get how to make a skip only with `tarantool.ConnectionPool` for `3.6`

test/suites/test_push.py
GRISHNOV: Done

test/suites/test_push.py
GRISHNOV: Done, deleted

test/suites/test_push.py
GRISHNOV: Done, left unchanged

test/suites/test_push.py
GRISHNOV: Done

test/suites/test_push.py
GRISHNOV: Done, deleted

test/suites/test_push.py
GRISHNOV: Done, deleted

test/suites/test_push.py
GRISHNOV: Done, implemented test cases inside `push_test_cases`

test/suites/lib/skip.py
```diff
@@ -162,3 +162,4 @@ def skip_or_run_datetime_test(func):
 
     return skip_or_run_test_pcall_require(func, 'datetime',
                                       'does not support datetime type')
+
```
DifferentialOrange, line 165: I'm not sure it's necessary. There is already a newline at the end of file.

tarantool/connection.py
```diff
@@ -766,14 +768,20 @@ def _read_response(self):
...
         Send request without trying to reconnect.
         Reload schema, if required.
 
         :param request: Request to send.
         :type request: :class:`~tarantool.request.Request`
 
+        :param on_push: Сallback for processing out-of-band messages.
+        :type on_push: :obj:`function`, optional
+
+        :param on_push_ctx: Сontext for working with on_push callback.
```
DifferentialOrange, line 782: Btw, you can use 
```
:paramref:`~tarantool.Connection._send_request_wo_reconnect.params.on_push`
```
here, but it's up to you.

docs/source/quick-start.rst
```diff
@@ -200,3 +200,151 @@ read-write and read-only pool instances:
...
+    callback_res = []
+
+    conn_pool = tarantool.ConnectionPool(
+            [{'host':'localhost', 'port':3301}],
+            user='guest')
+
+    res = conn_pool.insert(
+        'tester',
+        (1, 'Mike'),
+        mode=tarantool.Mode.ANY,
```
DifferentialOrange, line 332: I still think we should use the default one here, insert on `ANY` is also weird.

test/suites/test_push.py
DifferentialOrange: I think you may try `if hasattr(tarantool, 'ConnectionPool')` for this one. One more advice: if ConnectionPool is not supported, better set `self.conn_pool = None` and then check `if self.conn_pool is not None` on teardown, because, for some reason, `hasattr(self, 'conn_pool')` not works well in tests.

You may also wait for https://github.com/tarantool/tarantool-python/pull/245 (I'm not sure if it's closer to merging than your PR since I don't know how's the progress on https://github.com/tarantool/tarantool-python/pull/247#discussion_r1003155010) since it also contains backport of ConnectionPool to 3.6 and you don't need any skips anymore.

DifferentialOrange: By the way, I prefer using `tarantool.Connection` instead of `tarantool.connect` and thinking of deprecating the latter one. It would be nice if you've used it for consistency with modern tests.

test/suites/test_push.py
DifferentialOrange: Why push tests on windows platform are not supported? I don't see any restrictions from both Tarantool and connector side.

test/suites/test_push.py
```diff
@@ -0,0 +1,449 @@
...
+                    'tester',
+                    1,
+                    mode=tarantool.Mode.PREFER_RO,
+                    on_push=push_callback,
+                    on_push_ctx=callback_res
+                )
+            self.assertEqual(callback_res, [[[100, 1]], [[200, 1]], [[300, 1]]])
+            self.assertEqual(resp.data[0], [1, 'Bill'])
+
+        conn_pool = tarantool.ConnectionPool([{'host':self.host, 'port':self.port}], user='test', password='test')
```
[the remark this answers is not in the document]
DifferentialOrange: Sorry, it seems that my comment wasn't clear since actually some changes are required here. Let's add tests for MeshConnection

DifferentialOrange:
Sorry, it seems that you'll need to rebase one more time. I think it's would be the last time and your PR would be merged next.

Master changes that are relevant to you: now ConnectionPool is supported on Python 3.6 and you don't need to skip tests anymore.

test/suites/test_push.py
DifferentialOrange: Yeah, you don't need it anymore, see https://github.com/tarantool/tarantool-python/pull/247#issuecomment-1296784359

test/suites/test_push.py
GRISHNOV: Done, deleted this skip

test/suites/lib/skip.py
GRISHNOV: Done

docs/source/quick-start.rst
GRISHNOV: Done, now `ANY` is used only with `call` and `eval`

tarantool/connection.py
GRISHNOV: Thanks! Left unchanged

tarantool/connection.py
GRISHNOV: Sorry, but I couldn't reproduce the bug when changing the schema on the server in the process of receiving push:

I was making a delay in the client callback `callback(data, on_push_ctx=[])` with `time.sleep()`, and at that moment I was changing the scheme on the server with `box.space.tester:format()`, but there were no errors during push receiving

GRISHNOV:
Thank you for your feedback! I tried to answer all the comments on the code review

tarantool/connection.py
DifferentialOrange: To be honest, I also can't reproduce this one. I think we may just omit the testing for now and implement some adequate processing just in case it actually may happen.

After giving it some thought, I decided that Oleg's advice "If SchemaReloadException was thrown in the middle, just let it through" sounds like a good one. So I think we should use something like this
```python
response = None
while True:
    try:
        self._socket.sendall(bytes(request))
        response = request.response_class(self, self._read_response())
        break
    except SchemaReloadException as e:
        self.update_schema(e.schema_version)
        continue

while response._code == IPROTO_CHUNK:
    if on_push is not None: 
        if callable(on_push):
            on_push(response._data, on_push_ctx)
        else:
            raise TypeError('The on_push callback must be callable')
    response = request.response_class(self, self._read_response())

return response
```

and it should be alright if one manages to really get schema reload in the middle of the process.

test/suites/test_push.py
```diff
@@ -0,0 +1,472 @@
...
+                'server_function',
+                on_push=push_callback, 
+                on_push_ctx=callback_res
+            )
+            case = self.push_test_cases[assert_push_call.__name__]
+            self.assertEqual(callback_res, case['callback_res'])
+            self.assertEqual(resp.data[0], case['resp'])
+
+        # Check push from instance.
+        assert_push_call()
```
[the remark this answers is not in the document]
GRISHNOV: Done

test/suites/test_push.py
```diff
@@ -0,0 +1,472 @@
...
+    def setUp(self):
+        # Open connection, connection pool and mesh connection to instance.
+        self.conn = tarantool.Connection(host=self.host, port=self.port, 
+                                         user='test', password='test')
+        self.conn_pool = tarantool.ConnectionPool([{'host':self.host, 'port':self.port}], 
+                                                     user='test', password='test')
+        self.mesh_conn = tarantool.MeshConnection(host=self.host, port=self.port, 
+                                                  user='test', password='test')
+
+    push_test_cases = {
```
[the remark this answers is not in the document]
GRISHNOV: Done, now the name of the testing function (`call`, `insert`, ...) is contained in `push_test_cases`, as well as the input and output data for it. The `subTest` is also used now

tarantool/connection.py
GRISHNOV: Thanks! Done

test/suites/test_push.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def test_00_01_push_eval_via_connection(self):
+        def assert_push_eval():
+            callback_res = []
+            resp = self.conn.eval(
+                'return server_function()',
+                on_push=push_callback, 
+                on_push_ctx=callback_res
+            )
+            case = self.push_test_cases[assert_push_eval.__name__]
```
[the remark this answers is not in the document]
GRISHNOV: Done,` __name__ ` is not used now

tarantool/connection.py
```diff
@@ -796,6 +804,14 @@ def _send_request_wo_reconnect(self, request):
                 self.update_schema(e.schema_version)
                 continue
 
+        while response._code == IPROTO_CHUNK:
+            if on_push is not None: 
+                if callable(on_push):
```
oleg-jukovec, line 809: It is better to check it once in the object constructor.

GRISHNOV: Done, moved `callable` check to methods `call`, `eval`, `insert`, ...

oleg-jukovec: Thank you. I like it. @DifferentialOrange , could you offer a better solution?
Anyway, @GRISHNOV , be free to resolve the conversation.

tarantool/connection.py
```diff
@@ -785,7 +793,7 @@ def _send_request_wo_reconnect(self, request):
         """
 
         assert isinstance(request, Request)
-
+
```
DifferentialOrange, line 796: This change is both excessive and unwanted

tarantool/connection.py
```diff
@@ -796,6 +804,11 @@ def _send_request_wo_reconnect(self, request):
                 self.update_schema(e.schema_version)
                 continue
 
+        while response._code == IPROTO_CHUNK:
+            if on_push is not None: 
+                on_push(response._data, on_push_ctx)
```
DifferentialOrange, line 809: I think we may return back the type check here just in case

DifferentialOrange: Oh, I see you've moved in to another place.

test/suites/test_push.py
```diff
@@ -0,0 +1,257 @@
...
+        for case_name in self.push_test_cases.keys():
+            with self.subTest(name=case_name):
+                callback_res = []
+                case = self.push_test_cases[case_name]
+                testing_function = getattr(self.conn_pool, case_name)
+                case['input']['kwargs']['on_push_ctx'] = callback_res
+                resp = testing_function(
+                    *case['input']['args'],
+                    **case['input']['kwargs'],
+                    mode=tarantool.Mode.ANY
```
DifferentialOrange, line 241: Let's set `RW` here. If one would change the connection pool to have more than one instance, this call would fail on insert/delete/etc with `ALL`

test/suites/test_push.py
```diff
@@ -0,0 +1,257 @@
...
+                callback_res = []
+                case = self.push_test_cases[case_name]
+                testing_function = getattr(self.conn, case_name)
+                case['input']['kwargs']['on_push_ctx'] = callback_res
+                resp = testing_function(
+                    *case['input']['args'],
+                    **case['input']['kwargs']
+                )
+                self.assertEqual(callback_res, case['output']['callback_res'])
+                if len(resp.data) > 0:
```
DifferentialOrange, line 213: I think it's better to rely on "test case resp is None" condition. Otherwise we may miss something

DifferentialOrange (commented):
Thank you for you updates. I think this is the last round of comments.

Please, add the CHANGELOG entry (sorry for not having a convenient PR checklist yet).

tarantool/connection.py
GRISHNOV: Done, deleted

test/suites/test_push.py
GRISHNOV: Done, now it is `RW`

test/suites/test_push.py
GRISHNOV: Done, now using `case['output']['resp'] is not None` check

tarantool/connection.py
DifferentialOrange: I think `callable` is fine (and I'm not sure that there is any better approach).

In common, I thing it would be nice to use modern Python [type hints](https://docs.python.org/3/library/typing.html) some day because for now ~90% of arguments do not have any type asserts at all.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Support iproto feature push (#201).
```
DifferentialOrange, line 10: Let's add to to the bottom.

DifferentialOrange (approved):

CHANGELOG.md
GRISHNOV: Done, moved to the bottom of `Unreleased/Added`

oleg-jukovec (approved):
Thank you!

PR #251, opened 2022-10-24
Fix unittest warnings

### LINKED ISSUE #250: Fix unittest warnings
There are a lot of ResourseWarning and DeprecationWarning messages on unittest run that was hidden by `python setup.py test` and revealed by install package tests that uses direct `python -m unittest discover -v` command. It is not just an issue of clean log output: it also may result in fail for warning capture tests: https://github.com/tarantool/tarantool-python/actions/runs/3297461689/jobs/5439859418

DifferentialOrange:
### api: make connection close idempotent

After this patch, it is possible to call `conn.close()` method multiple
times.

Part of #250

### test: replace deprecated methods

assertRaisesRegexp and assertEquals were renamed in version 3.2 [1].

1. https://docs.python.org/3/library/unittest.html#deprecated-aliases

Part of #250

### test: fix server resource warnings

Fix several ResourceWarning messages related to test Tarantool server
methods not cleaning up sockets.

Part of #250

### test: fix unix socket test resource warnings

hasattr check actually has no effect for unittest runtime-added fields,
so both server and connection for unix socket test weren't closed after
run.

Part of #250

### test: fix test connection resource warnings

Close all connection opened in tests even in case of failed asserts.

Part of #250

### test: fix dbapi test connection resource warnings

DBAPI2 compliance tests are not implemented here but inherited from
external module [1]. Two tests from this module open a connection and
forget to close it. The issue [2] had been fixed and PR had been merged,
but there is no tagged release yet.

1. https://pypi.org/project/dbapi-compliance/
2. https://github.com/baztian/dbapi-compliance/issues/5

Part of #250

### test: filter warnings for disabled instances

There are several ConnectionPool tests that stop some pool instances
and verify that everything works fine even for semi-functional cluster.
Different network and cluster state warning are issued in this case.
They are expected and not informative, thus it is better to filter them
in final output.

Closes #250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Package build (#238).
 - Allow any MessagePack supported type as a request key (#240).
 - Puting test files in pip package (#238).
+- Make connection close idempotent (#250).
 
 ## 0.9.0 - 2022-06-20
 
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -1,3 +1,3 @@
-dbapi-compliance==1.15.0
+git+https://github.com/baztian/dbapi-compliance.git@ea7cb1b4#egg=dbapi-compliance
 pyyaml==6.0
 importlib-metadata >= 1.0 ; python_version < '3.8'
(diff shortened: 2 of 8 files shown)


DifferentialOrange:
image
Oh, well, maybe it's worth to wait until a release

GRISHNOV (approved):
LGTM

DifferentialOrange:
> image Oh, well, maybe it's worth to wait until a release

Install package from master commit for tests

Issue #204, opened 2021-12-14
Support datetime


Totktonada:
It is not offered ATM (2021-12-14), but AFAIK there is agreement on the representation in msgpack. See https://github.com/tarantool/tarantool/pull/6660.

Python documentation: https://docs.python.org/3/library/datetime.html.

DifferentialOrange:
Changed estimation since now it includes timezones too.

PR #252, opened 2022-10-24
Single approach to ext types encode/decode

DifferentialOrange:
### api: pandas way to build datetime from timestamp

This option is required so it would be possible to decode Datetime with
external function without constructing excessive pandas.Timestamp
object.

Follows #204

### api: extract datetime encode/decode from class

Extract tarantool.Datetime encode and decode to external functions. This
is a breaking change, but since there is no tagged release with Datetime
yet and API was more internal rather than public, it shouldn't be an
issue.

Follows #204

### api: extract interval encode/decode from class

Extract tarantool.Interval encode and decode to external functions. This
is a breaking change, but since there is no tagged release with Interval
yet and API was more internal rather than public, it shouldn't be an
issue.

Follows #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -136,6 +136,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support iproto feature discovery (#206).
 
+- Support pandas way to build datetime from timestamp (PR #252).
+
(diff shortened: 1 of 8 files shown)


oleg-jukovec (approved):
Thank you, LGTM!

PR #263, opened 2022-11-11
internal: drop graceful shutdown feature flag

### LINKED ISSUE #262: Drop graceful shutdown feature flag
`IPROTO_FEATURE_GRACEFUL_SHUTDOWN` feature flag was dropped in https://github.com/tarantool/tarantool/commit/2e9cbec3091e7c5d6ca1d7dad4305333e70f5876 commit before release, so there is no traces of this flag in Tarantool releases and it is possible that some new meaning would be used for `IPROTO_FEATURE_X = 4`, so we must drop it too to not confuse users.

DifferentialOrange:
Graceful shutdown was introduced in Tarantool with [1]. A week later it was reworked with [2]. So the timeline is as follows.

- Graceful shutdown is introduced, new feature flag is added and the protocol version is bumped to 4.
- Graceful shutdown is reworked, feature flag is dropped and the protocol version is dropped to 3.
- Next Tarantool version released.

So the idea for users is "IPROTO_FEATURE_GRACEFUL_SHUTDOWN and protocol version 4 never existed". It also means that any new feature flag may have value `4` (same as IPROTO_FEATURE_GRACEFUL_SHUTDOWN), but with completely different meaning. IPROTO_FEATURE_GRACEFUL_SHUTDOWN flag was added in [3] by mistake due to misunderstanding the event timeline. This patch removes the flag. Since `_features` dict is an internal of a connection object, the change shouldn't break anything for users.

1. https://github.com/tarantool/tarantool/commit/6f29f9d7725331d74d5e9a57a4922b1c3801ad50
2. https://github.com/tarantool/tarantool/commit/2e9cbec3091e7c5d6ca1d7dad4305333e70f5876
3. https://github.com/tarantool/tarantool-python/pull/243

Closes #262


GRISHNOV (approved):
Hi! LGTM

PR #264, opened 2022-11-15
api: native crud module support

### LINKED ISSUE #205: Convenient API for tarantool/crud
There is https://github.com/tarantool/crud/ module, which provides easy API to access data in sharded cluster. We can use it from the connector (using `<connection object>.call(<...>)`), but it would be convenient to have a kind of native API for tarantool/crud.

GRISHNOV:
Adds native api support for crud module to use it from a connection object.

Below there are examples of using the api. 

```python
>>> import tarantool
>>> from tarantool.error import CrudModuleError, CrudModuleManyError, DatabaseError
>>> conn = tarantool.Connection(host='localhost',port=3301)

>>> conn.crud_
    conn.crud_count(                conn.crud_insert(               conn.crud_insert_object_many(   
    conn.crud_min(                  conn.crud_replace_object(       conn.crud_stats(                
    conn.crud_unflatten_rows(       conn.crud_upsert_many(          conn.crud_delete(               
    conn.crud_insert_many(          conn.crud_len(                  conn.crud_replace(              
    conn.crud_replace_object_many(  conn.crud_storage_info(         conn.crud_update(               
    conn.crud_upsert_object(        conn.crud_get(                  conn.crud_insert_object(        
    conn.crud_max(                  conn.crud_replace_many(         conn.crud_select(               
    conn.crud_truncate(             conn.crud_upsert(               conn.crud_upsert_object_many(
```

### Inserting data via crud:

```python
# Insert without exception:
>>> res = conn.crud_insert('tester', (3500,300,'Rob'))
>>> res
<tarantool.crud.CrudResult object at 0x11a56e320>
>>> res.
res.metadata  res.rows
>>> res.rows
[[3500, 300, 'Rob']]
>>> res.metadata
[{'name': 'id', 'type': 'unsigned'}, {'name': 'bucket_id', 'type': 'unsigned'}, {'name': 'name', 'type': 'string'}]

# Insert with exception (duplicate key exists):
>>> try:
...     res = conn.crud_insert('tester', (3500,300,'Rob'))
... except CrudModuleError as e:
...     exc_crud = e
... 
>>> exc_crud
CrudModuleError(0, 'Failed to insert: Duplicate key exists in unique index "primary_index" in space "tester" with old tuple - [3500, 300, "Rob"] and new tuple - [3500, 300, "Rob"]')
>>> exc_crud.extra_info_error
<tarantool.crud.CrudError object at 0x10a276950>
>>> exc_crud.extra_info_error.
exc_crud.extra_info_error.class_name  exc_crud.extra_info_error.err         exc_crud.extra_info_error.file        exc_crud.extra_info_error.line        exc_crud.extra_info_error.str       
>>> exc_crud.extra_info_error.class_name
'InsertError'
>>> exc_crud.extra_info_error.str
'InsertError: Failed to insert: Duplicate key exists in unique index "primary_index" in space "tester" with old tuple - [3500, 300, "Rob"] and new tuple - [3500, 300, "Rob"]'

# In case of batch operation (*_many), CrudModuleManyError exception contains both result and errors (if there is a problem with at least one row).
>>> try:
...     res = conn.crud_insert_object_many('tester', ({'id':3,'bucket_id':100,'name':'Ann'}, {'id':4,'bucket_id':100,'name':'Sam'}), {'timeout':100, 'rollback_on_error':False})
... except CrudModuleManyError as e:
...     exc_crud = e
... 
>>> exc_crud
CrudModuleManyError(0, 'Got multiple errors, see errors_list')
>>> exc_crud.success_list # some of the rows were inserted.
<tarantool.crud.CrudResult object at 0x11a56f310>
>>> exc_crud.success_list.rows
[[1, 100, 'Bob'], [2, 100, 'Rob']]
>>> exc_crud.errors_list # some of the rows were not inserted.
[<tarantool.crud.CrudError object at 0x11a56e9e0>, <tarantool.crud.CrudError object at 0x11a56f490>]
>>> exc_crud.errors_list[0].str
'CallError: Failed for 037adb3a-b9e3-4f78-a6d1-9f0cdb6cbefc: Function returned an error: Duplicate key exists in unique index "primary_index" in space "tester" with old tuple - [3500, 300, "Rob"] and new tuple - [3500, 100, "Mike"]'
>>> exc_crud.errors_list[1].str
'InsertManyError: Failed to flatten object: FlattenError: Object is specified in bad format: FlattenError: Unknown field "second_name" is specified'
    
# If there are no problems with any rows, the entire response will be contained in the res variable.
>>> res = conn.crud_insert_object_many('tester', ({'id':3,'bucket_id':100,'name':'Ann'}, {'id':4,'bucket_id':100,'name':'Sam'}), {'timeout':100, 'rollback_on_error':False})
>>> res.rows
[[3, 100, 'Ann'], [4, 100, 'Sam']]
```

### Crud not found on the router:
```python
>>> try:
...     res = conn.crud_insert('tester', (22221,300,'Rob'))
... except DatabaseError as e:
...     exc_db = e
... 
 >>> exc_db
DatabaseError(33, "Ensure that you're calling crud.router and user has sufficient grants")
>>> exc_db.extra_info
BoxError(type='ClientError', file='/tmp/tarantool-20221003-6335-edruh3/tarantool-2.10.3/src/box/lua/call.c', line=112, message="Procedure 'crud.insert' is not defined", errno=0, errcode=33, fields=None, prev=None)
```

### Select and unflatten_rows via crud:

```python
>>> try:
...     res = conn.crud_select('tester', {}, {'first':2})
... except CrudModuleError as e:
...     exc_crud = e
... 
>>> res
<tarantool.crud.CrudResult object at 0x10a276d10>
>>> res.
res.metadata  res.rows      
>>> res.rows
[[1, 100, 'Mike'], [2, 100, 'Mike']]
>>> res.metadata
[{'name': 'id', 'type': 'unsigned'}, {'name': 'bucket_id', 'type': 'unsigned'}, {'name': 'name', 'type': 'string'}]
>>> r = conn.crud_unflatten_rows(res.rows, res.metadata)
>>> r
[{'id': 1, 'bucket_id': 100, 'name': 'Mike'}, {'id': 2, 'bucket_id': 100, 'name': 'Mike'}]
```

### Truncate and len via crud:

```python
>>> try:
...     res = conn.crud_len('tester')
... except CrudModuleError as e:
...     exc_crud = e
>>> res
26
>>> try:
...     res = conn.crud_truncate('tester')
... except CrudModuleError as e:
...     exc_crud = e
>>> res
True
```

### Encoding=None in Connection:

```python
>>> conn = tarantool.Connection(host='localhost',port=3301,encoding=None)

>>> try:
...     res = conn.crud_insert('tester', (3500,300,'Rob'))
... except CrudModuleError as e:
...     exc_crud = e
... 
>>> res
<tarantool.crud.CrudResult object at 0x1188a7fd0>
>>> res.
res.metadata  res.rows      
>>> res.metadata
[{b'name': b'id', b'type': b'unsigned'}, {b'name': b'bucket_id', b'type': b'unsigned'}, {b'name': b'name', b'type': b'string'}]
>>> res.rows
[[3500, 300, b'Rob']]
```

Closes #205

### DIFF
--- a/.github/workflows/packing.yml
+++ b/.github/workflows/packing.yml
@@ -99,6 +99,12 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
 
+      - name: Install the crud module for testing purposes
+        run: |
+          curl -L https://tarantool.io/release/2/installer.sh | bash
+          sudo apt install -y tt
+          tt rocks install crud
+
       - name: Run tests
         run: make test-pure-install
 
@@ -331,6 +337,12 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
 
+      - name: Install the crud module for testing purposes
+        run: |
+          curl -L https://tarantool.io/release/2/installer.sh | bash
+          sudo apt install -y tt
+          tt rocks install crud
+
       - name: Run tests
         run: make test-pure-install
 
@@ -491,6 +503,12 @@ jobs:
 
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
+
+      - name: Install the crud module for testing purposes
+        run: |
+          curl -L https://tarantool.io/release/2/installer.sh | bash
+          sudo apt install -y tt
+          tt rocks install crud
 
       - name: Run tests
         run: make test-pure-install
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -85,6 +85,12 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
 
+      - name: Install the crud module for testing purposes
+        run: |
+          curl -L https://tarantool.io/release/2/installer.sh | bash
+          sudo apt install -y tt
+          tt rocks install crud
+
       - name: Run tests
         run: make test
 
@@ -143,6 +149,15 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
 
+      - name: Install the crud module for testing purposes
+        # This is a workaround with TARANTOOL_DIR and should be reworked later.
+        # See more here: https://github.com/tarantool/tt/issues/282
+        run: |
+          source tarantool-enterprise/env.sh
+          curl -L https://tarantool.io/release/2/installer.sh | bash
+          sudo apt install -y tt
+          tt rocks install crud TARANTOOL_DIR=$PWD/tarantool-enterprise
+
       - name: Run tests
         run: |
           source tarantool-enterprise/env.sh
@@ -195,6 +210,12 @@ jobs:
 
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
+
+      - name: Install the crud module for testing purposes
+        run: |
+          curl -L https://tarantool.io/release/2/installer.sh | bash
+          sudo apt install -y tt
+          tt rocks install crud
 
       - name: Run tests
         run: make test-pure-install
--- a/docs/source/quick-start.rst
+++ b/docs/source/quick-start.rst
@@ -347,3 +347,123 @@ with out-of-band message processing:
 
     print(callback_res)
     >>> [[[100, 1]], [[200, 1]], [[300, 1]]]
+
+
+Interaction with the crud module
+----------------------------------
+
+Through the :class:`~tarantool.Connection` object, you can access 
+`crud module <https://github.com/tarantool/crud>`_ methods:
+
+.. code-block:: python
+
+    >>> import tarantool
+    >>> from tarantool.error import CrudModuleError, CrudModuleManyError, DatabaseError
+    >>> conn = tarantool.Connection(host='localhost',port=3301)
+
+    >>> conn.crud_
+    conn.crud_count(                conn.crud_insert(               conn.crud_insert_object_many(   
+    conn.crud_min(                  conn.crud_replace_object(       conn.crud_stats(                
+    conn.crud_unflatten_rows(       conn.crud_upsert_many(          conn.crud_delete(               
+    conn.crud_insert_many(          conn.crud_len(                  conn.crud_replace(              
+    conn.crud_replace_object_many(  conn.crud_storage_info(         conn.crud_update(               
+    conn.crud_upsert_object(        conn.crud_get(                  conn.crud_insert_object(        
+    conn.crud_max(                  conn.crud_replace_many(         conn.crud_select(               
+    conn.crud_truncate(             conn.crud_upsert(               conn.crud_upsert_object_many(
+
+As an example, consider :meth:`~tarantool.Connection.crud_insert` and :meth:`~tarantool.Connection.crud_insert_object_many`. 
+It is recommended to enclose calls in the try-except construction as follows:
+
+.. code-block:: python
+
+    # Insert without exception:
+    >>> res = conn.crud_insert('tester', (3500,300,'Rob'))
+    >>> res
+    <tarantool.crud.CrudResult object at 0x11a56e320>
+    >>> res.
+    res.metadata  res.rows
+    >>> res.rows
+    [[3500, 300, 'Rob']]
+    >>> res.metadata
+    [{'name': 'id', 'type': 'unsigned'}, {'name': 'bucket_id', 'type': 'unsigned'}, {'name': 'name', 'type': 'string'}]
+
+    # Insert with exception (duplicate key exists):
+    >>> try:
+    ...     res = conn.crud_insert('tester', (3500,300,'Rob'))
+    ... except CrudModuleError as e:
+    ...     exc_crud = e
+    ... 
+    >>> exc_crud
+    CrudModuleError(0, 'Failed to insert: Duplicate key exists in unique index "primary_index" in space "tester" with old tuple - [3500, 300, "Rob"] and new tuple - [3500, 300, "Rob"]')
+    >>> exc_crud.extra_info_error
+    <tarantool.crud.CrudError object at 0x10a276950>
+    >>> exc_crud.extra_info_error.
+    exc_crud.extra_info_error.class_name  exc_crud.extra_info_error.err         exc_crud.extra_info_error.file        exc_crud.extra_info_error.line        exc_crud.extra_info_error.str       
+    >>> exc_crud.extra_info_error.class_name
+    'InsertError'
+    >>> exc_crud.extra_info_error.str
+    'InsertError: Failed to insert: Duplicate key exists in unique index "primary_index" in space "tester" with old tuple - [3500, 300, "Rob"] and new tuple - [3500, 300, "Rob"]'
+
+    # In case of batch operation (*_many), CrudModuleManyError exception contains both result and errors (if there is a problem with at least one row).
+    >>> try:
+    ...     res = conn.crud_insert_object_many('tester', ({'id':3,'bucket_id':100,'name':'Ann'}, {'id':4,'bucket_id':100,'name':'Sam'}), {'timeout':100, 'rollback_on_error':False})
+    ... except CrudModuleManyError as e:
+    ...     exc_crud = e
+    ... 
+    >>> exc_crud
+    CrudModuleManyError(0, 'Got multiple errors, see errors_list')
+    >>> exc_crud.success_list # some of the rows were inserted.
+    <tarantool.crud.CrudResult object at 0x11a56f310>
+    >>> exc_crud.success_list.rows
+    [[1, 100, 'Bob'], [2, 100, 'Rob']]
+    >>> exc_crud.errors_list # some of the rows were not inserted.
+    [<tarantool.crud.CrudError object at 0x11a56e9e0>, <tarantool.crud.CrudError object at 0x11a56f490>]
+    >>> exc_crud.errors_list[0].str
+    'CallError: Failed for 037adb3a-b9e3-4f78-a6d1-9f0cdb6cbefc: Function returned an error: Duplicate key exists in unique index "primary_index" in space "tester" with old tuple - [3500, 300, "Rob"] and new tuple - [3500, 100, "Mike"]'
+    >>> exc_crud.errors_list[1].str
+    'InsertManyError: Failed to flatten object: FlattenError: Object is specified in bad format: FlattenError: Unknown field "second_name" is specified'
+    
+    # If there are no problems with any rows, the entire response will be contained in the res variable.
+    >>> res = conn.crud_insert_object_many('tester', ({'id':3,'bucket_id':100,'name':'Ann'}, {'id':4,'bucket_id':100,'name':'Sam'}), {'timeout':100, 'rollback_on_error':False})
+    >>> res.rows
+    [[3, 100, 'Ann'], [4, 100, 'Sam']]
+
+If module crud not found on the router or user has not sufficient grants:
+
+.. code-block:: python
+
+    >>> try:
+    ...     res = conn.crud_insert('tester', (22221,300,'Rob'))
+    ... except DatabaseError as e:
+    ...     exc_db = e
+    ... 
+    >>> exc_db
+    DatabaseError(33, "Procedure 'crud.insert' is not defined. Ensure that you're calling crud.router and user has sufficient grants")
+    >>> exc_db.extra_info
+    BoxError(type='ClientError', file='/tmp/tarantool-20221003-6335-edruh3/tarantool-2.10.3/src/box/lua/call.c', line=112, message="Procedure 'crud.insert' is not defined", errno=0, errcode=33, fields=None, prev=None)
+
+Using :meth:`~tarantool.Connection.crud_select` and :meth:`~tarantool.Connection.crud_unflatten_rows`:
+
+.. code-block:: python
+
+    >>> res = conn.crud_select('tester', [], {'first':2})
+    >>> res
+    <tarantool.crud.CrudResult object at 0x10a276d10>
+    >>> res.rows
+    [[1, 100, 'Mike'], [2, 100, 'Bill']]
+    >>> res.metadata
+    [{'name': 'id', 'type': 'unsigned'}, {'name': 'bucket_id', 'type': 'unsigned'}, {'name': 'name', 'type': 'string'}]
+    >>> r = conn.crud_unflatten_rows(res.rows, res.metadata)
+    >>> r
+    [{'id': 1, 'bucket_id': 100, 'name': 'Mike'}, {'id': 2, 'bucket_id': 100, 'name': 'Bill'}]
+
+Using :meth:`~tarantool.Connection.crud_truncate` and :meth:`~tarantool.Connection.crud_len`:
+
+.. code-block:: python
+
+    >>> res = conn.crud_len('tester')
+    >>> res
+    26
+    >>> res = conn.crud_truncate('tester')
+    >>> res
+    True
--- a/tarantool/connection.py
+++ b/tarantool/connection.py
@@ -84,6 +84,8 @@ from tarantool.error import (
     InternalError,
     ProgrammingError,
     NotSupportedError,
+    CrudModuleError,
+    CrudModuleManyError,
     SchemaReloadException,
     Warning,
     warn
@@ -95,6 +97,12 @@ from tarantool.utils import (
     wrap_key,
     ENCODING_DEFAULT,
 )
+from tarantool.crud import (
+    CrudResult,
+    CrudError,
+    call_crud,
+)
+from typing import Union
 
 # Based on https://realpython.com/python-interface/
 class ConnectionInterface(metaclass=abc.ABCMeta):
@@ -244,6 +252,198 @@ class ConnectionInterface(metaclass=abc.ABCMeta):
 
         raise NotImplementedError
 
+    @abc.abstractmethod
+    def crud_insert(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_insert`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_insert_object(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_insert_object`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_insert_many(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_insert_many`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_insert_object_many(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_insert_object_many`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_get(self, space_name, key, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_get`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_update(self, space_name, key, operations=[], opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_update`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_delete(self, space_name, key, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_delete`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_replace(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_replace`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_replace_object(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_replace_object`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_replace_many(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_replace_many`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_replace_object_many(self, space_name, values, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_replace_object_many`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_upsert(self, space_name, values, operations=[], opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_upsert`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_upsert_object(self, space_name, values, operations=[], opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_upsert_object`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_upsert_many(self, space_name, values_operation, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_upsert_many`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_upsert_object_many(self, space_name, values_operation, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_upsert_object_many`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_select(self, space_name: str, conditions=[], opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_select`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_min(self, space_name, index_name, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_min`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_max(self, space_name, index_name, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_max`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_truncate(self, space_name, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_truncate`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_len(self, space_name, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_len`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_storage_info(self, opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_storage_info`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_count(self, space_name, conditions=[], opts={}):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_count`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_stats(self, space_name=None):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_stats`.
+        """
+
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def crud_unflatten_rows(self, rows, metadata):
+        """
+        Reference implementation: :meth:`~tarantool.Connection.crud_unflatten_rows`.
+        """
+
+        raise NotImplementedError
+
 
 class Connection(ConnectionInterface):
     """
@@ -344,6 +544,13 @@ class Connection(ConnectionInterface):
     DBAPI compatibility.
 
     :value: :exc:`~tarantool.error.NotSupportedError`
+    """
+
+    CrudModuleError = CrudModuleError
+    """
+    DBAPI compatibility.
+
+    :value: :exc:`~tarantool.error.CrudModuleError`
     """
 
     def __init__(self, host, port,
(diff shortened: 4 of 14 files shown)


tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+
+    def __init__(self, con):
+        """
+        TODO: write docs
+        """
+
+        self.con = con
+        self._crud_module_init_complete = False
+
+    def _crud_module_init(self):
```
DifferentialOrange, line 40: https://github.com/tarantool/crud/blob/704bd6f1f1d6683906d800935450ef971ca4ccac/crud.lua#L176-L178

It seems that simply enabling crud router exposes crud handles to global. I've tried to experiment on it by myself and seems fine.

Moreover, doing explicit `crud = require('crud')` create global `crud` variable on storages (and in real clusters storages are mostly not routers) so this make this check more confusing (you can't call anything on storages).

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+    Contains crud module methods.
+    """
+
+    def __init__(self, con):
+        """
+        TODO: write docs
+        """
+
+        self.con = con
+        self._crud_module_init_complete = False
```
DifferentialOrange, line 38: As far as I understand, there are two main reasons for `Crud` class:
- to store `_crud_module_init_complete` variable,
- to be able to use module as a namespace of `conn`: `conn.crud`, since you'll need conn anyway.

I think it is possible to get rid of both of them.

Above I've discussed that there are seems to be no need to call `crud = require('crud')`. You may check that the instance is a crud router (we may call any operations on routers only, we have no interest in interacting with storages explicitly) by `rawget(_G, 'crud') != nil`. But do we really need that? It seems like an optimization, but it is rather hard to have our "cache" consistent: if one would disable crud router (highly unlikely, but not impossible), it will show that require is ok which is untrue. And the main point is we actually don't need to know if we're able to call crud procedures: if we're don't, we will get "Procedure not defined" (code 33) error. It is rather simple to make this error more useful: just wrap each code 33 error with comment "Ensure that you're calling crud.router and user have enough grants". It seems no use to fast fail since if user would try to call `crud` thousands of times from a connection to an instance without `crud` they are doing something really weird and it's not worth to optimize it.

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+        """
+
+        if not self._crud_module_init_complete:
+            self._crud_module_init()
+        assert self.require_ok == True, 'Problem with require of crud module on the remote machine'
+        assert isinstance(space_name, str)
+        assert isinstance(values, (tuple, list))
+        assert isinstance(opts, dict)
+
+        crud_resp = self.con.call("crud.insert", space_name, values, opts)
```
DifferentialOrange, line 71: We also need to think a bit about grants. Using `eval` requires universe grants or something really close to that. Using `call`s requires dramatically less grants. If we can reduce our code to using only calls then the user which are concerned about grants hierarchy will be able to create a user with grants to call several specific functions and it will work.

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+
+        crud_resp = self.con.call("crud.stats", space_name)
+
+        res = None
+        if len(crud_resp.data[0]) > 0:
+            res = Result(crud_resp.data[0])
+
+        return res
+
+    def reset_stats(self):
```
DifferentialOrange, line 676: I think we also shouldn't provide `reset_stats` and `cfg` handles. Sure, it isn't hard for us to expose them, but I think that instance configuration is something that should be performed on server side. If one really would want to perform such things, they may always mess with call and eval.

On the other hand, having `stats` call shouldn't hurt anyone and we may leave it.

DifferentialOrange (changes requested):
Thank you for your draft! The task it much trickier that I initially have expected. I gave it some thought (and recalled some `crud` behavior) and I think that we should try to rework some things. I'll summarize my comments here.
- Try to get rid of class `Crud`.
- Try to work with `call`s only.
- Do not expose configuration handles.
- Raise errors as exceptions.
- Optimize utils usage.

I see that the first one is a requirement to be able to use `conn.crud.method`, but is such approach really widespread in Python? I'm not well-experienced myself, but I think that I have seen providing many handles from a single place more often than using complex hierarchy. Maybe it's worth to study this question a little by using some Python modules as an example (and if you still decide to use `conn.crud.method`, maybe you'll see some nice implementation approaches).

It is highly likely that I have forgot to mention something important in this review, so feel free to ask me any additional questions here or in personal chat.

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
+class Response(object):
+    def __init__(self, response):
+        """
+        TODO: write docs
+        """
+
+        if isinstance(response, dict):
+            for response_field_name in response.keys():
+                setattr(self, response_field_name, response[response_field_name])
+        elif isinstance(response, str):
```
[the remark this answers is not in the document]
GRISHNOV: Done, wrote an example in the description of PR

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+    def __init__(self, con):
+        """
+        TODO: write docs
+        """
+
+        self.con = con
+        self._crud_module_init_complete = False
+
+    def _crud_module_init(self):
+        self.require_ok = self.con.eval("ok, crud = pcall(require, 'crud') return ok").data[0]
```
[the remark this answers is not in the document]
GRISHNOV: Done, now the `require` call for `crud` is not being made

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+
+    def upsert_many(self, space_name, values_operation, opts={}):
+        """
+        TODO: write docs
+        """
+
+        if not self._crud_module_init_complete:
+            self._crud_module_init()
+        assert self.require_ok == True, 'Problem with require of crud module on the remote machine'
+        assert isinstance(space_name, str)
```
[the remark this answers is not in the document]
GRISHNOV: Done, wrote annotations to parameters and return values of crud functions

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+        
+        res = None
+        if crud_resp[0] is not None:
+            res = Result(crud_resp[0])
+
+        err = None
+        if crud_resp[1] is not None:
+            err = Error(crud_resp[1])
+
+        return res, err
```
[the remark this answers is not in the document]
GRISHNOV: Done, now work with errors through exceptions, updated PR description

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+
+    def pairs(self, space_name, conditions={}, opts={}):
```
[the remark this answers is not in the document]
GRISHNOV: At the moment I don't quite understand how to implement `pairs`, I'm working on it now

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+        """
+
+        if not self._crud_module_init_complete:
+            self._crud_module_init()
+        assert self.require_ok == True, 'Problem with require of crud module on the remote machine'
+        crud_resp = self.con.call("crud.reset_stats")
+
+        return crud_resp.data[0]
+
+    def init_router(self):
```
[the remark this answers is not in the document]
GRISHNOV: Done, deleted

tarantool/crud.py
```diff
@@ -0,0 +1,748 @@
...
+        """
+        TODO: write docs
+        """
+
+        if not self._crud_module_init_complete:
+            self._crud_module_init()
+        assert self.require_ok == True, 'Problem with require of crud module on the remote machine'
+        self.con.call("crud.stop_storage")
+
+    def unflatten_rows(self, rows, metadata):
```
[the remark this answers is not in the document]
GRISHNOV: Done, implemented locally, an example of the `unflatten` work added to the description of PR

tarantool/crud.py
GRISHNOV: Done, now the require call for crud is not being made

tarantool/crud.py
GRISHNOV: Done, integrated functions directly into Connection (conn.crud_*). Also, when an `DatabaseError` exception with code 33 appears, I supplement it with auxiliary information

tarantool/crud.py
GRISHNOV: Done, now the `eval` is not used, but it may be required during implementation `pairs` (I'm working on it now)

tarantool/crud.py
GRISHNOV: Done, deleted `reset_stats` and `cfg`

GRISHNOV:
Thank you for your feedback! 

I have made most of the edits (with the exception of the `pairs` implementation).
If the current API is suitable, I'll move on to writing documentation and tests

tarantool/crud.py
Totktonada: Related issues:

* https://github.com/tarantool/crud/issues/229
* https://github.com/tarantool/cartridge-java/issues/159

DifferentialOrange:
> ```python
> >>> exc_crud.res is None # use only in case of *_many
> ```

I don't get this comment.

GRISHNOV:
> ```python
> >>> exc_crud.res is None # use only in case of *_many
> ```
> I don't get this comment.

I mean, in exc_crud.res, the result can only appear when the `*_many()` is called (when part of the batch is successful and part is not), in other cases exc_crud.res will always be None

tarantool/connection.py
```diff
@@ -1751,3 +1765,563 @@ def _check_features(self):
...
+
+    def crud_get(self, space_name: str, key: int, opts={}) -> CrudResult:
+        """
+        TODO: write docs
+        """
+
+        assert isinstance(space_name, str)
+        assert isinstance(opts, dict)
+
+        try:
```
DifferentialOrange, line 1875: It seems like you can extract this block to a single wrapper.

tarantool/error.py
```diff
@@ -330,6 +330,28 @@ class PoolTolopogyError(DatabaseError):
     pass
 
 
+class CrudModuleError(Error):
```
DifferentialOrange, line 333: I think it should inherit DatabaseError

tarantool/connection.py
```diff
@@ -343,6 +350,13 @@ class Connection(ConnectionInterface):
     :value: :exc:`~tarantool.error.NotSupportedError`
     """
 
+    CrudModuleError = CrudModuleError
+    """
+    DBAPI compatibility.
```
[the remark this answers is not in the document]
Totktonada: However it is convenient to access an exceptions using a field of a class or an instance without explicit import.

tarantool/crud.py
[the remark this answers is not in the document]
GRISHNOV: At the moment, it has been decided not to implement `pairs`. I suppose it's resolved

tarantool/crud.py
GRISHNOV: At the moment, it has been decided not to implement `pairs`. I suppose it's resolved

tarantool/connection.py
```diff
@@ -1751,3 +1765,563 @@ def _check_features(self):
         features_list = [val for val in CONNECTOR_FEATURES if val in server_features]
         for val in features_list:
             self._features[val] = True
+
+    def crud_insert(self, space_name: str, values: Union[tuple, list], opts={}) -> CrudResult:
+        """
+        TODO: write docs
+        """
+
+        assert isinstance(space_name, str)
```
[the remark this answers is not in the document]
GRISHNOV: Done, AFAIU, changes are not required here

tarantool/connection.py
GRISHNOV: Done, implemented `call_crud` in tarantool.crud

tarantool/error.py
GRISHNOV: Done

tarantool/connection.py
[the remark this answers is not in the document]
DifferentialOrange: On the other hand, having too much things exposed from the single point is a bit confusing. To be honest, I'm not sure which approach is the best for us. So it's up to you for now.

GRISHNOV:
Thank you! 
At the moment, I have implemented tests and documentation, and removed the draft status from the PR

tarantool/connection.py
```diff
@@ -1751,3 +1765,765 @@ def _check_features(self):
         features_list = [val for val in CONNECTOR_FEATURES if val in server_features]
         for val in features_list:
             self._features[val] = True
+
+    def crud_insert(self, space_name: str, values: Union[tuple, list], opts={}) -> CrudResult:
+        """
+        Inserts row through the `tarantool-crud <https://github.com/tarantool/crud#insert>`__ module.
```
DifferentialOrange, line 1771: > tarantool-crud

That's definitely not the way to name it. It's either `tarantool/crud` or simply `crud`. Since we have a link here, I prefet the latter one: `crud`.

DifferentialOrange, line 1771: I'm sorry that we don't have a linter for now.

PEP8 [declares](https://peps.python.org/pep-0008/#maximum-line-length) that
> For flowing long blocks of text with fewer structural restrictions (docstrings or comments), the line length should be limited to 72 characters.

This restriction **includes** the tabulation before docstings. There are many places in this package where this rule is violated, but in all of them we simply can't do it any other way.

For example, you can solve the issue here by moving the link in a separate line, like this
https://github.com/tarantool/tarantool-python/blob/70b0a3debe5ffa7f1c0b5c84d6fb9e28cbe26152/tarantool/connection.py#L469

tarantool/connection.py
```diff
@@ -1751,3 +1765,765 @@ def _check_features(self):
...
+
+        :param space_name: The name of the target space.
+        :type space_name: :obj:`str`
+
+        :param values: The data for the crud module.
```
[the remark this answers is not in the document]
DifferentialOrange: The same about other methods.

tarantool/connection.py
```diff
@@ -1751,3 +1765,765 @@ def _check_features(self):
...
+            :exc:`~tarantool.error.DatabaseError`
+        """
+
+        assert isinstance(space_name, str)
+        assert isinstance(values, (tuple, list))
+        assert isinstance(opts, dict)
+
+        crud_resp = call_crud(self, "crud.insert", space_name, values, opts)
+
+        if crud_resp[1] is not None:
```
DifferentialOrange, line 1794: Why re-raising hasn't been put to a `call_crud` method? You may use a couple of them if you need to (like `call_` and `call_many_`

tarantool/crud.py
```diff
@@ -0,0 +1,81 @@
+"""
+This module provides API for interaction with a crud module.
```
DifferentialOrange, line 2: At first, it's definitely "the" crud module.

And we definitely need a link here.

tarantool/crud.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+
+from tarantool.error import DatabaseError
+
+
+class CrudResponse(object):
+    """
+    Contains response fields from the `crud`_ module that correspond to the lua implementation.
```
DifferentialOrange, line 10: 72 characters rule.

tarantool/crud.py
```diff
@@ -0,0 +1,81 @@
...
+                else:
+                    setattr(self, response_field_name, response[response_field_name])
+        elif isinstance(response, str) or isinstance(response, bytes):
+            self.str = response
+            self.err = response
+        else:
+            raise RuntimeError('Unable to decode response to object due to unknown type')
+
+
+class CrudResult(CrudResponse):
```
DifferentialOrange, line 36: I don't see any reason to implement `CrudResponse` class which serves two different purposes through if/else condition in init. Just put corresponding init parts to `CrudResult` and `CrudError`.

test/suites/crud_server.lua
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env tarantool
+
+os = require('os')
```
DifferentialOrange, line 3: `local`?

test/suites/crud_server_cfg.lua
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env tarantool
```
DifferentialOrange, line 1: Is it a duplicate of `crud_server.lua`?

test/suites/test_crud.py
```diff
@@ -0,0 +1,584 @@
...
+
+def create_server():
+    srv = TarantoolServer()
+    srv.script = 'test/suites/crud_server.lua'
+    srv.start()
+
+    return srv
+
+
+class TestSuite_Crud(unittest.TestCase):
```
DifferentialOrange, line 18: To actually run a test, you need to specify your file here:

https://github.com/tarantool/tarantool-python/blob/70b0a3debe5ffa7f1c0b5c84d6fb9e28cbe26152/test/suites/__init__.py#L26-L32

And after you do it, CI run will fail cause you never install `crud` module in the pipeline. So you need to fix it too.

test/suites/test_crud.py
```diff
@@ -0,0 +1,584 @@
...
+        time.sleep(1)
+        # Open connection to instance.
+        self.conn = tarantool.Connection(host='localhost', port=self.port, 
+                                         user='guest', password='')
+        # Time for vshard group configuration.
+        time.sleep(1)
+
+    crud_test_cases = {
+        'crud_insert': {
+            'correct_input': {
```
DifferentialOrange, line 38: Having too nested structure is a pain, but I think that input and output should be inside the "success"/"error" cases.

test/suites/test_crud.py
```diff
@@ -0,0 +1,584 @@
...
+                        *case['correct_input']['args'],
+                    )
+                    if 'rows' in case['correct_output']:
+                        self.assertEqual(resp.rows, case['correct_output']['rows'])
+                    if 'scalar' in case['correct_output']:
+                        # Case for scalar value as crud responce, not tarantool.crud.CrudResult obj.
+                        self.assertEqual(resp, case['correct_output']['scalar'])
+                except CrudModuleError as e:
+                    self.assertEqual(e, None)
+                # Exception try.
```
DifferentialOrange, line 555: Since it is already the two different code blocks which test the different behavior, let's make two separate dicts with cases and two separate test methods.

test/suites/test_crud.py
```diff
@@ -0,0 +1,584 @@
...
+                    ],
+                    [
+                        {'name': 'id', 'type': 'unsigned'}, 
+                        {'name': 'bucket_id', 'type': 'unsigned'}, 
+                        {'name': 'name', 'type': 'string'}
+                    ],
+                ],
+            },
+            'correct_output': {
+                'scalar': [
```
DifferentialOrange, line 520: What is `scalar`? I don't remember such fields in crud.

test/suites/test_crud.py
```diff
@@ -0,0 +1,584 @@
...
+            'exception_input': {
+                'args': [[],[]],
+            },
+            'exception_output': {
+                'str': [],
+            },
+        },
+    }
+
+    def test_crud_module_via_connection(self):
```
DifferentialOrange, line 538: You should also run the same tests for connection_pool and mesh_connection (if the latter would be too time-consuming, you may omit it).

test/suites/test_crud.py
```diff
@@ -0,0 +1,584 @@
...
+
+    def setUp(self):
+        time.sleep(1)
+        # Open connection to instance.
+        self.conn = tarantool.Connection(host='localhost', port=self.port, 
+                                         user='guest', password='')
+        # Time for vshard group configuration.
+        time.sleep(1)
+
+    crud_test_cases = {
```
DifferentialOrange, line 36: I don't see `stats` and `truncate` here. Since we provide these handles and their behavior is specific in nature,we definitely should test them.

test/suites/test_crud.py
```diff
@@ -0,0 +1,584 @@
...
+        print(' CRUD '.center(70, '='), file=sys.stderr)
+        print('-' * 70, file=sys.stderr)
+        # Create server and extract helpful fields for tests.
+        self.srv = create_server()
+        self.port = self.srv.args['primary']
+
+    def setUp(self):
+        time.sleep(1)
+        # Open connection to instance.
+        self.conn = tarantool.Connection(host='localhost', port=self.port,
```
DifferentialOrange, line 31: Shouldn't we extract the host the same way we extract the port? Moreover, localhost is failed to resolve in docker: https://github.com/tarantool/tarantool-python/commit/70adf7d91bda23aadbb08d914fa5047309dfbbec

tarantool/connection.py
```diff
@@ -81,6 +81,7 @@
     InternalError,
```
DifferentialOrange, line 81: https://github.com/tarantool/tarantool-python/blob/70b0a3debe5ffa7f1c0b5c84d6fb9e28cbe26152/tarantool/connection.py#L96-L241

I think that crud native API should become the part of `Connection` API.

tarantool/connection.py
```diff
@@ -81,6 +81,7 @@
...
     ProgrammingError,
```
DifferentialOrange, line 82: Moreover, it doesn't seems like connection_pool supports crud native API now. Choosing default modes may be a bit complicated, so feel free to ask if needed. Maybe we should provide a comment like "crud native API in pool makes sense only for pool consisting of crud routers" too.

DifferentialOrange (commented):
Thank you for the update!

See comments below. The main drawback of the current version is that tests are actually not run on CI, so I think you would be interested to fix it first and then fix everything that fails on CI.

There are some thing that I may haven't covered yet, but there should be enough work for now.

docs/source/quick-start.rst
```diff
@@ -347,3 +347,154 @@ with out-of-band message processing:
...
+    conn.crud_count(                conn.crud_insert(               conn.crud_insert_object_many(   
+    conn.crud_min(                  conn.crud_replace_object(       conn.crud_stats(                
+    conn.crud_unflatten_rows(       conn.crud_upsert_many(          conn.crud_delete(               
+    conn.crud_insert_many(          conn.crud_len(                  conn.crud_replace(              
+    conn.crud_replace_object_many(  conn.crud_storage_info(         conn.crud_update(               
+    conn.crud_upsert_object(        conn.crud_get(                  conn.crud_insert_object(        
+    conn.crud_max(                  conn.crud_replace_many(         conn.crud_select(               
+    conn.crud_truncate(             conn.crud_upsert(               conn.crud_upsert_object_many(
+
+As an example, consider :class:`~tarantool.Connection.crud_insert` and :class:`~tarantool.Connection.crud_insert_object_many`.
```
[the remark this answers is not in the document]
GRISHNOV: Done

docs/source/quick-start.rst
```diff
@@ -347,3 +347,154 @@ with out-of-band message processing:
...
+    conn.crud_upsert_object(        conn.crud_get(                  conn.crud_insert_object(        
+    conn.crud_max(                  conn.crud_replace_many(         conn.crud_select(               
+    conn.crud_truncate(             conn.crud_upsert(               conn.crud_upsert_object_many(
+
+As an example, consider :class:`~tarantool.Connection.crud_insert` and :class:`~tarantool.Connection.crud_insert_object_many`. 
+It is recommended to enclose calls in the try-except construction as follows:
+
+.. code-block:: python
+
+    # Try without exception:
```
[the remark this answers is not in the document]
GRISHNOV: Done

docs/source/quick-start.rst
```diff
@@ -347,3 +347,154 @@ with out-of-band message processing:
...
+    >>> res
+    <tarantool.crud.CrudResult object at 0x11a56e320>
+    >>> res.
+    res.metadata  res.rows # fields like in lua implementation.
+    >>> res.rows
+    [[3500, 300, 'Rob']]
+    >>> res.metadata
+    [{'name': 'id', 'type': 'unsigned'}, {'name': 'bucket_id', 'type': 'unsigned'}, {'name': 'name', 'type': 'string'}]
+
+    # Try with exception:
```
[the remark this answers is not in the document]
GRISHNOV: Done

docs/source/quick-start.rst
```diff
@@ -347,3 +347,154 @@ with out-of-band message processing:
...
+    ... 
+    >>> exc_crud
+    CrudModuleError(<tarantool.crud.CrudResult object at 0x11a56f310>, [<tarantool.crud.CrudError object at 0x11a56e9e0>, <tarantool.crud.CrudError object at 0x11a56f490>])
+    >>> exc_crud.
+    exc_crud.args             exc_crud.errs             exc_crud.res              exc_crud.with_traceback(  
+    >>> exc_crud.res # some of the lines were inserted.
+    <tarantool.crud.CrudResult object at 0x11a56f310>
+    >>> exc_crud.res.rows
+    [[1800, 100, 'Mike'], [3800, 100, 'Bob']]
+    >>> exc_crud.errs # some of the lines were not inserted.
```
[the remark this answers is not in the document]
GRISHNOV: Done

docs/source/quick-start.rst
```diff
@@ -347,3 +347,154 @@ with out-of-band message processing:
...
+    >>> try:
+    ...     res = conn.crud_insert('tester', (22221,300,'Rob'))
+    ... except CrudModuleError as e:
+    ...     exc_crud = e
+    ... except DatabaseError as e:
+    ...     exc_db = e
+    ... 
+    >>> exc_db
+    DatabaseError(33, "Procedure 'crud.insert' is not defined. Ensure that you're calling crud.router and user have enough grants")
+    >>> exc_db.
```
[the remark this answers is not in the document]
GRISHNOV: Done

tarantool/connection.py
```diff
@@ -1751,3 +1765,765 @@ def _check_features(self):
         features_list = [val for val in CONNECTOR_FEATURES if val in server_features]
         for val in features_list:
             self._features[val] = True
+
+    def crud_insert(self, space_name: str, values: Union[tuple, list], opts={}) -> CrudResult:
```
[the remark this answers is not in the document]

tarantool/connection.py

tarantool/connection.py
GRISHNOV: I thought about it, but there are methods in the Lua implementation that give not 2 values, but one. I noted this behavior with comments in the code `<In absence of an error, crud does not give variable err as nil (as in most cases)>`. Also, in some methods, not a table is returned, but a scalar value (for example, boolean, like `truncate`)

AFAIU, it is difficult to implement it just through `call_` and `call_many_`

tarantool/crud.py
GRISHNOV: Done

tarantool/crud.py
GRISHNOV: Done

test/suites/crud_server.lua
GRISHNOV: Done

test/suites/crud_server_cfg.lua
GRISHNOV: Done, deleted

test/suites/test_crud.py
GRISHNOV: Done. At the moment I have problem with implementing CI tests under windows. AFAIU, there is a problem with the [availability of methods](https://github.com/tarantool/tarantool-python/actions/runs/3628331837/jobs/6119215806#step:11:1957), the reason for which I cannot yet determine. The rest of the cases are working properly

test/suites/test_crud.py
GRISHNOV: Done

test/suites/test_crud.py
GRISHNOV: Done

test/suites/test_crud.py
GRISHNOV: Done, now implemented tests for `mesh connection` and `connection pool`

test/suites/test_crud.py
GRISHNOV: Done, added

test/suites/test_crud.py
GRISHNOV: Done, now it works with 0.0.0.0

tarantool/connection.py
GRISHNOV: Done, added

tarantool/connection.py
GRISHNOV: Done. AFAIU, I implemented test cases for `mesh connection` and `connection pool`, and now it works

test/suites/crud_server.lua
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env tarantool
+
+local os = require('os')
```
DifferentialOrange, line 3: `require('os')` is excessive since `os` is exposed to global in Tarantool

```
$ tarantool
Tarantool 2.10.0-beta2-283-gb56e012bc
type 'help' for interactive help
tarantool> os
---
- execute: 'function: builtin#116'
  rename: 'function: builtin#118'
  environ: 'function: 0x40be3f60'
  setenv: 'function: 0x402c8458'
  setlocale: 'function: builtin#126'
  getenv: 'function: builtin#120'
  difftime: 'function: builtin#125'
  remove: 'function: builtin#117'
  date: 'function: builtin#123'
  exit: 'function: 0x40bdd868'
  time: 'function: builtin#124'
  clock: 'function: builtin#122'
  tmpname: 'function: builtin#119'
...
```

DifferentialOrange:
> For unknown reasons, your PR doesn't trigger `packing` pipeline. I'll check it out.

[Doc](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request):
```
Note: Workflows will not run on pull_request activity if the pull request has a merge conflict.
The merge conflict must be resolved first.
```

It would be nice if you've resolved them first so CI runs would be full (`packing` has test runs too).

tarantool/crud.py
DifferentialOrange: I mean something like this

```
class CrudResult():
    def __init__(self, response):
        assert(response, dict)

        for response_field_name in response.keys():
            if isinstance(response_field_name, bytes):
                setattr(self, response_field_name.decode(), response[response_field_name])
            else:
                setattr(self, response_field_name, response[response_field_name])

class CrudError():
    def __init__(self, response):
        assert isinstance(response, (str, bytes))
        self.str = response
        self.err = response
```

since we never create `CrudResponse` objects: it is always either a `CrudResult` or a `CrudError`

tarantool/crud.py
```diff
@@ -0,0 +1,87 @@
...
+def wrapCrudUndefinedProcedureError(e: DatabaseError):
+    """
+    Supplements information about an unsuccessful attempt 
+    to find the crud module procedure.
+
+    :param e: The error object of the crud module call.
+    :type e: :class:`~tarantool.error.DatabaseError`
+    """
+
+    info = ". Ensure that you're calling crud.router and user have enough grants"
```
DifferentialOrange, line 62:
```suggestion
    info = ". Ensure that you're calling crud.router and user has sufficient grants"
```

tarantool/crud.py
```diff
@@ -0,0 +1,87 @@
...
+    :type conn: :class:`~tarantool.connection.Connection`
+
+    :param args: The arguments to pass to the crud module.
+    :type args: :obj:`tuple`
+    """
+
+    try:
+        crud_resp = conn.call(*args)
+    except DatabaseError as e:
+        if e.code == 33:
```
DifferentialOrange, line 83: Let's use a named constant here instead of `33` (unfortunately, current `error.py` is a bit messy)

test/suites/test_crud.py
```diff
@@ -0,0 +1,751 @@
...
+        self.conn_pool = tarantool.ConnectionPool([{'host':self.host, 'port':self.port}], 
+                                                     user='guest', password='')
+        # Time for vshard group configuration.
+        time.sleep(1)
+
+    crud_test_cases = {
+        'crud_insert': {
+            'success': {
+                'correct_input': {
+                'args': ['tester', [1, 100, 'Mike'], {'timeout': 10}],
```
DifferentialOrange, line 45: Indentation is weird here.

test/suites/test_crud.py
```diff
@@ -0,0 +1,751 @@
...
+                                         user='guest', password='')
+        self.conn_pool = tarantool.ConnectionPool([{'host':self.host, 'port':self.port}], 
+                                                     user='guest', password='')
+        # Time for vshard group configuration.
+        time.sleep(1)
+
+    crud_test_cases = {
+        'crud_insert': {
+            'success': {
+                'correct_input': {
```
DifferentialOrange, line 44: Since we're already in `success`, why not use `input`/`output`? Same for error.

test/suites/test_crud.py
```diff
@@ -0,0 +1,751 @@
...
+            if 'scalar' in case['success']['correct_output']:
+                # Case for scalar value as crud responce, not tarantool.crud.CrudResult obj.
+                self.assertEqual(resp, case['success']['correct_output']['scalar'])
+            if 'operations' in case['success']['correct_output']:
+                # Case for statistics testing.
+                for operation in case['success']['correct_output']['operations']:
+                    self.assertEqual(operation in resp.spaces['tester'], True, 
+                        'Problem with finding a field with a statistic about operation ' 
+                            + operation)
+        except CrudModuleError as e:
```
DifferentialOrange, line 681: I think it's okay to let CrudModuleError just fall through and finish the test case with error.

tarantool/connection.py
```diff
@@ -1751,3 +1957,771 @@ def _check_features(self):
...
+
+        return crud_resp[0]
+
+    def crud_stats(self, space_name=None) -> CrudResult:
+        """
+        Gets statistics through the 
+        `crud <https://github.com/tarantool/crud#statistics>`__.
+
+        :param space_name: The name of the target space.
+        :type space_name: :obj:`str`
```
DifferentialOrange, line 2685:
```suggestion
        :type space_name: :obj:`str`,Optional
```

You're also missing type hint.

tarantool/connection.py
```diff
@@ -1751,3 +1957,771 @@ def _check_features(self):
...
+        crud_resp = call_crud(self, "crud.count", space_name, conditions, opts)
+
+        if crud_resp[1] is not None:
+            raise CrudModuleError(None, CrudError(crud_resp[1]))
+
+        return crud_resp[0]
+
+    def crud_stats(self, space_name=None) -> CrudResult:
+        """
+        Gets statistics through the
```
DifferentialOrange, line 2681:
```suggestion
        Gets statistics from the 
```

test/suites/test_crud.py
```diff
@@ -0,0 +1,751 @@
...
+                    'str': [
+                            r'"no-such-space-name" doesn\'t exist', 
+                        ],
+                },
+            },
+        },
+        'crud_stats': {
+            'success': {
+                'correct_input': {
+                    'args': [],
```
DifferentialOrange, line 634: Let's add one more test with space name. Maybe it'll be convenient to extract specific handles from this test dict to some separate one (up to you).

DifferentialOrange (commented):
Thank you for the updates. I'll give it one last shot after you fix the CI and resolve remaining comments. `tarantool_python_ci.lua` may be a bit complicated so feel free to ask anything.

tarantool/crud.py
GRISHNOV: Unfortunately, the code above does not give the desired behavior. 
I can suggest such code for consideration, but it will duplicate some behavior in init:

```python
class CrudResult():
    """
    Contains result's fields from result variable 
    of crud module operation.
    """

    def __init__(self, response):
        """
        Sets response fields as in Lua implementation.

        :param response: The response object of the crud module call.
        :type response: :class:`~tarantool.response.Response`
        """

        if isinstance(response, dict):
            for response_field_name in response.keys():
                if isinstance(response_field_name, bytes):
                    setattr(self, response_field_name.decode(), response[response_field_name])
                else:
                    setattr(self, response_field_name, response[response_field_name])


class CrudError():
    """
    Contains error's fields from error variable 
    of crud module operation.
    """

    def __init__(self, response):
        """
        Sets response fields as in Lua implementation.

        :param response: The response object of the crud module call.
        :type response: :class:`~tarantool.response.Response`
        """

        if isinstance(response, dict):
            for response_field_name in response.keys():
                if isinstance(response_field_name, bytes):
                    setattr(self, response_field_name.decode(), response[response_field_name])
                else:
                    setattr(self, response_field_name, response[response_field_name])
        elif isinstance(response, str) or isinstance(response, bytes):
            self.str = response
            self.err = response
```

test/suites/crud_server.lua
GRISHNOV: Done, deleted

.github/workflows/packing.yml
```diff
@@ -162,6 +165,13 @@ jobs:
           touch tarantool.pid
           echo $TNT_PID > ./tarantool.pid
 
+      - name: Install the crud module for testing purposes
+        shell: wsl-bash_Ubuntu-20.04 {0}
+        run: |
+          sudo apt install -y unzip
+          tarantoolctl rocks install crud
+          tarantoolctl rocks list
```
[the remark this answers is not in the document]
GRISHNOV: Done, deleted

tarantool/connection.py
```diff
@@ -1751,3 +1957,771 @@ def _check_features(self):
...
+        :type opts: :obj:`dict`, optional
+
+        :rtype: :class:`~tarantool.crud.CrudResult`
+
+        :raise: :exc:`~tarantool.error.CrudModuleError`,
+            :exc:`~tarantool.error.DatabaseError`
+        """
+
+        assert isinstance(space_name, str)
+        assert isinstance(values, dict)
```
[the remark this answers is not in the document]
GRISHNOV: Done

tarantool/crud.py
GRISHNOV: Done

test/suites/test_crud.py
GRISHNOV: Done

test/suites/test_crud.py
GRISHNOV: Done

test/suites/test_crud.py
GRISHNOV: Done

tarantool/connection.py

tarantool/connection.py

test/suites/test_crud.py
GRISHNOV: Done, now this test case passes the name of space to the testing function

tarantool/crud.py
GRISHNOV: Yep, it seems that the case was not needed:

```python
elif isinstance(response, str) or isinstance(response, bytes):
        self.str = response
        self.err = response
```

Then version of code proposed by me for consideration will look like this:

```python
class CrudResult():
    """
    Contains result's fields from result variable 
    of crud module operation.
    """

    def __init__(self, response):
        """
        Sets response fields as in Lua implementation.

        :param response: The response object of the crud module call.
        :type response: :class:`~tarantool.response.Response`
        """

        if isinstance(response, dict):
            for response_field_name in response.keys():
                if isinstance(response_field_name, bytes):
                    setattr(self, response_field_name.decode(), response[response_field_name])
                else:
                    setattr(self, response_field_name, response[response_field_name])


class CrudError():
    """
    Contains error's fields from error variable 
    of crud module operation.
    """

    def __init__(self, response):
        """
        Sets response fields as in Lua implementation.

        :param response: The response object of the crud module call.
        :type response: :class:`~tarantool.response.Response`
        """

        if isinstance(response, dict):
            for response_field_name in response.keys():
                if isinstance(response_field_name, bytes):
                    setattr(self, response_field_name.decode(), response[response_field_name])
                else:
                    setattr(self, response_field_name, response[response_field_name])
```

DifferentialOrange: Well, in this case it really makes sense to use a single class with inheritance.

test/suites/test_crud.py
DifferentialOrange: Seems it is complicated to configure vshard replicasets under WSL so let's skip this one.

tarantool/error.py
```diff
@@ -330,6 +330,33 @@ class PoolTolopogyError(DatabaseError):
     pass
 
 
+class CrudModuleError(DatabaseError):
```
DifferentialOrange, line 333: This one does not implement `DatabaseError` in any way. I think the simplest way here is to inherit `Error`.

tarantool/error.py
```diff
@@ -330,6 +330,33 @@ class PoolTolopogyError(DatabaseError):
...
+        # Sets tarantool.crud.CrudResult object.
+        self.res = args[0]
+
+        if isinstance(args[1], (tuple, list)):
+            # Case for crud_*_many().
+            # Sets list of tarantool.crud.CrudError objects.
+            self.errs = args[1]
+        else:
+            # Sets tarantool.crud.CrudError object.
+            self.err = args[1]
```
DifferentialOrange, line 357: Which fields do we have in CrudError? Can't we simply expose them here so it be similar to existing errors?

tarantool/crud.py
```diff
@@ -0,0 +1,87 @@
...
+
+    :param conn: The connection object for the crud module call.
+    :type conn: :class:`~tarantool.connection.Connection`
+
+    :param args: The arguments to pass to the crud module.
+    :type args: :obj:`tuple`
+    """
+
+    # Response error code for case "Requested procedure is not defined".
+    ER_NO_SUCH_PROC = 33
```
DifferentialOrange, line 78: Let's put it in `const.py` or `error.py`. At least let's not create the constant on any function invocation.

tarantool/connection_pool.py
```diff
@@ -1025,3 +1025,638 @@ def execute(self, query, params=None, *, mode=None):
...
+
+        :rtype: :class:`~tarantool.crud.CrudResult`
+
+        :raise: :exc:`~tarantool.error.CrudModuleError`,
+            :exc:`~tarantool.error.DatabaseError`
+        """
+
+        return self._send(mode, 'crud_replace_object_many', space_name, values, opts)
+
+    def crud_upsert(self, space_name, values, operations=[], opts={}, *, mode=Mode.RW):
```
DifferentialOrange, line 1329: Since `crud_*` always deal with routers and `crud.router`s doesn't write anything on their own, `Mode.ANY` should be better here (and in any other call).

docs/source/quick-start.rst
```diff
@@ -347,3 +347,140 @@ with out-of-band message processing:
...
+    >>> try:
+    ...     res = conn.crud_insert('tester', (3500,300,'Rob'))
+    ... except CrudModuleError as e:
+    ...     exc_crud = e
+    ... 
+    >>> exc_crud
+    CrudModuleError(None, <tarantool.crud.CrudError object at 0x10a276950>)
+    >>> exc_crud.
+    exc_crud.args             exc_crud.err              exc_crud.res              exc_crud.with_traceback(  
+    >>> exc_crud.args
```
DifferentialOrange, line 400: `args` is an internal, we shouldn't discuss this one.

docs/source/quick-start.rst
```diff
@@ -347,3 +347,140 @@ with out-of-band message processing:
...
+    # Insert with exception (duplicate key exists):
+    >>> try:
+    ...     res = conn.crud_insert('tester', (3500,300,'Rob'))
+    ... except CrudModuleError as e:
+    ...     exc_crud = e
+    ... 
+    >>> exc_crud
+    CrudModuleError(None, <tarantool.crud.CrudError object at 0x10a276950>)
+    >>> exc_crud.
+    exc_crud.args             exc_crud.err              exc_crud.res              exc_crud.with_traceback(
```
DifferentialOrange, line 399: Let's remove this line too.

docs/source/quick-start.rst
```diff
@@ -347,3 +347,140 @@ with out-of-band message processing:
...
+    >>> exc_crud.args
+    (<tarantool.crud.CrudResult object at 0x11a56f310>, [<tarantool.crud.CrudError object at 0x11a56e9e0>, <tarantool.crud.CrudError object at 0x11a56f490>])
+
+If module crud not found on the router:
+
+.. code-block:: python
+
+    >>> try:
+    ...     res = conn.crud_insert('tester', (22221,300,'Rob'))
+    ... except CrudModuleError as e:
```
DifferentialOrange, line 442: It's excessive

docs/source/quick-start.rst
```diff
@@ -347,3 +347,140 @@ with out-of-band message processing:
...
+    ...     exc_db = e
+    ... 
+    >>> exc_db
+    DatabaseError(33, "Procedure 'crud.insert' is not defined. Ensure that you're calling crud.router and user have enough grants")
+
+Using :meth:`~tarantool.Connection.crud_select` and :meth:`~tarantool.Connection.crud_unflatten_rows`:
+
+.. code-block:: python
+
+    >>> try:
```
DifferentialOrange, line 454: Again, I don't think that it is the Python way to wrap any function call in try/catch. Let's not write in such a way until you've want to work with exception as an object.

docs/source/quick-start.rst
```diff
@@ -347,3 +347,140 @@ with out-of-band message processing:
...
+    [{'name': 'id', 'type': 'unsigned'}, {'name': 'bucket_id', 'type': 'unsigned'}, {'name': 'name', 'type': 'string'}]
+    >>> r = conn.crud_unflatten_rows(res.rows, res.metadata)
+    >>> r
+    [{'id': 1, 'bucket_id': 100, 'name': 'Mike'}, {'id': 2, 'bucket_id': 100, 'name': 'Mike'}]
+
+Using :meth:`~tarantool.Connection.crud_truncate` and :meth:`~tarantool.Connection.crud_len`:
+
+.. code-block:: python
+
+    >>> try:
```
DifferentialOrange, line 475: And here too

DifferentialOrange (commented):
I think the only one unresolved question is errors. Every other thing seems ok.

tarantool/error.py
GRISHNOV: Done

tarantool/crud.py
GRISHNOV: Done

tarantool/connection_pool.py
GRISHNOV: Done

docs/source/quick-start.rst
GRISHNOV: Done

docs/source/quick-start.rst
GRISHNOV: Done

docs/source/quick-start.rst
GRISHNOV: Done

docs/source/quick-start.rst
GRISHNOV: Done

docs/source/quick-start.rst
GRISHNOV: Done

tarantool/error.py
GRISHNOV: It depends on the method we are calling. At the moment we are just exposing the fields from the Lua implementation (`args[0]` is `res` and `args[1]` is `err/errs` Lua vars). We can implement something like switch-case for each method, but is it necessary? Or did I misunderstand the question

DifferentialOrange: Let's mimic a `DatabaseError` in a simple way.

### What's in crud

#### For non-many
We have `res, err`, where `err` is a single error table from `errors` rock and `res` is `nil`.

#### For many
We have `res, err`, where `err` is an array of table errors and `res` is common result like in plain return.

### What's in python

`DatabaseError` is an `Exception` with `code`, `message` and `extra_info` (the same error, but with additional fields based on provided `box.error` info).

### How we can mix it

First case is rather simple. We can raise some error with the same code and message we've got from Tarantool. But since we've got additional info, it would be nice to provide it.

Second case is much complicated. My proposal is as follows:

- Have two different classes for non-many and many errors (`CrudModuleError` and `CrudModuleManyError`).
- Both should inherit `DatabaseError`.

For `CrudModuleError`, let's add the `extra_info_errors` attribute with the original error (`DatabaseError` has `extra_info` field with similar meaning, but different structure).

For `CrudModuleManyError`, let's raise error based on `DatabaseError(0, "Got multiple errors, see errors_list")`, store original errors list in `errors_list` attribute and success tuples in `success_tuples` attribute (I'm not insist on naming here -- it's just an example, but I think the current `res`, `errs` one is a bit confusing).

You also may define a new class for `errors` Lua rock errors since they always have the same structure, but this is optional -- dict is fine too.

tarantool/crud.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class CrudError(CrudResponse):
+    """
+    Contains error's fields from error variable 
+    of crud module operation.
+    """
+
+
+def wrapCrudUndefinedProcedureError(e: DatabaseError):
```
DifferentialOrange, line 50: First, let's create a new error instead of messing with existing one. I think chaining mechanism may be really useful here: https://docs.python.org/3/tutorial/errors.html#exception-chaining

tarantool/crud.py
```diff
@@ -0,0 +1,84 @@
...
+    to find the crud module procedure.
+
+    :param e: The error object of the crud module call.
+    :type e: :class:`~tarantool.error.DatabaseError`
+    """
+
+    info = ". Ensure that you're calling crud.router and user has sufficient grants"
+    e.message = e.message + info
+    e.extra_info.message = e.extra_info.message + info
+    e.args = (e.args[0], e.args[1] + info)
```
DifferentialOrange, line 62: Second, let's never mess with `args`.

```
>>> class MyError(Exception):
...     def __init__(self, *args):
...             pass
... 
>>> MyError(1, 2)
MyError(1, 2)
>>> e = MyError(1, 2)
>>> e.
e.args             e.with_traceback(  
>>> e.args
(1, 2)
```

`args` exception attribute is essentially a copy of "what data was used to raise exception with".

tarantool/crud.py
```diff
@@ -0,0 +1,84 @@
...
+    return e
+
+def call_crud(conn, *args):
+    """
+    Calls the crud via connection.call with try/except block.
+
+    :param conn: The connection object for the crud module call.
+    :type conn: :class:`~tarantool.connection.Connection`
+
+    :param args: The arguments to pass to the crud module.
```
DifferentialOrange, line 73: Actually, it's not: it is crud method and the arguments to pass to the crud module

tarantool/crud.py
DifferentialOrange, line 50: Python codestyle is to use camel_case to name functions.

DifferentialOrange, line 50: I think both methods here should have `:meta private:` in docstrings: they are internal and I don't see any reason a user may need to call them explicitly.

tarantool/crud.py
```diff
@@ -0,0 +1,84 @@
...
+    :type args: :obj:`tuple`
+    """
```
DifferentialOrange, line 75: Btw, it's better to clarify `:raises:` here, see `connection.py` for an example.

tarantool/error.py
GRISHNOV: done

.github/workflows/packing.yml
```diff
@@ -99,6 +99,12 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
 
+      - name: Install the crud module for testing purposes
+        run: |
+          curl -L https://tarantool.io/KOzQfax/release/2/installer.sh | bash
```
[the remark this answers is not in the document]
GRISHNOV: done

.github/workflows/testing.yml
```diff
@@ -143,6 +149,13 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
 
+      - name: Install the crud module for testing purposes
+        run: |
+          source tarantool-enterprise/env.sh
+          curl -L https://tarantool.io/KOzQfax/release/2/installer.sh | bash
+          sudo apt install -y tt
+          tt rocks install crud TARANTOOL_DIR=/home/runner/work/tarantool-python/tarantool-python/tarantool-enterprise
```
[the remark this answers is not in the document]
GRISHNOV: done

tarantool/crud.py
GRISHNOV: done

tarantool/crud.py
GRISHNOV: done

GRISHNOV: done

tarantool/crud.py
GRISHNOV: done

tarantool/error.py
GRISHNOV: Done, I also updated the documentation in `quick-start.rst`

tarantool/crud.py
GRISHNOV: Done, also added a case with error handling with code 42 (`ER_ACCESS_DENIED`). As far as I understand, now the exception chaining works as it should

tarantool/crud.py
GRISHNOV: Done, deleted

.github/workflows/testing.yml
```diff
@@ -143,6 +149,14 @@ jobs:
       - name: Install test requirements
         run: pip3 install -r requirements-test.txt
 
+      - name: Install the crud module for testing purposes
+        # This is a workaround with TARANTOOL_DIR and should be reworked later.
```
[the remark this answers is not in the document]
GRISHNOV: Done, now there is a link to tt#282

tarantool/error.py
```diff
@@ -330,6 +330,52 @@ class PoolTolopogyError(DatabaseError):
...
 
+class CrudModuleError(DatabaseError):
+    """
+    Exception raised for errors that are related to
+    the operation result of `crud`_ module.
+
+    .. _crud: https://github.com/tarantool/crud
+    """
+
+    def __init__(self, *args):
```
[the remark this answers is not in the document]
GRISHNOV: Done, now using `__init__(self, success, error):`

tarantool/error.py
```diff
@@ -330,6 +330,52 @@ class PoolTolopogyError(DatabaseError):
...
+        """
+        Sets fields with result and errors.
+        
+        :param args: The tuple from the crud module with result and errors.
+        :type args: :obj:`tuple`
+        """
+
+        # Sets tarantool.crud.CrudResult object.
+        self.success = args[0]
```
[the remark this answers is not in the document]
GRISHNOV: Done, deleted

tarantool/connection_pool.py
```diff
@@ -1025,3 +1025,638 @@ def execute(self, query, params=None, *, mode=None):
             raise ValueError("Please, specify 'mode' keyword argument")
 
         return self._send(mode, 'execute', query, params)
+
+    def crud_insert(self, space_name, values, opts={}, *, mode=Mode.ANY):
+        """
+        Execute an CRUD_INSERT request on the pool server:
```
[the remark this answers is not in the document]
GRISHNOV: Done, rewritten without uppercase

GRISHNOV:
Thank you for your feedback! I have updated the PR description and tried to answer all the comments on the code review

tarantool/crud.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def call_crud(conn, *args):
+    """
+    Calls the crud via connection.call with try/except block.
+
+    :param conn: The connection object for the crud module call.
+    :type conn: :class:`~tarantool.connection.Connection`
+
+    :param args: The args for the crud method.
```
DifferentialOrange, line 57: method name + args for the crud method

tarantool/crud.py
```diff
@@ -0,0 +1,72 @@
...
+    :raise: :exc:`~tarantool.error.DatabaseError`
+
+    :meta private:
+    """
+
+    try:
+        crud_resp = conn.call(*args)
+    except DatabaseError as e:
+        if e.code == ER_NO_SUCH_PROC or e.code == ER_ACCESS_DENIED:
+            exc_msg = "Ensure that you're calling crud.router and user has sufficient grants"
```
DifferentialOrange, line 69: Let's do `e.msg + exc_msg`, as before

tarantool/error.py
```diff
@@ -330,6 +330,51 @@ class PoolTolopogyError(DatabaseError):
...
 
+class CrudModuleError(DatabaseError):
+    """
+    Exception raised for errors that are related to
+    the operation result of `crud`_ module.
+
+    .. _crud: https://github.com/tarantool/crud
+    """
+
+    def __init__(self, success, error):
```
DifferentialOrange, line 341:
```suggestion
    def __init__(self, _, error):
```

docs/source/quick-start.rst
```diff
@@ -347,3 +347,123 @@ with out-of-band message processing:
...
+    ...     res = conn.crud_insert('tester', (22221,300,'Rob'))
+    ... except DatabaseError as e:
+    ...     exc_db = e
+    ... 
+    >>> exc_db
+    DatabaseError(33, "Ensure that you're calling crud.router and user has sufficient grants")
+    >>> exc_db.extra_info
+    BoxError(type='ClientError', file='/tmp/tarantool-20221003-6335-edruh3/tarantool-2.10.3/src/box/lua/call.c', line=112, message="Procedure 'crud.insert' is not defined", errno=0, errcode=33, fields=None, prev=None)
```
DifferentialOrange, line 443: Don't forget to update here after message rework.

DifferentialOrange (approved):
Thank you for your updates! Let's resolve a couple of minor comments and I'll merge this one.

tarantool/crud.py
GRISHNOV: Done, now `The method name and args for the crud method.`

tarantool/crud.py
GRISHNOV: Done

tarantool/error.py
GRISHNOV: Done

docs/source/quick-start.rst
GRISHNOV: Updated

DifferentialOrange (approved):

PR #265, opened 2022-11-18
ci: use merge commit for pull_request_target

oleg-jukovec:
It would be more correct to use a merge request commit instead of original commit for pull_request_target.


oleg-jukovec:
```15 skipped, 25 successful, and 23 in progress checks```

Wow =)

DifferentialOrange (approved):
Thank you for your work!

PR #268, opened 2022-11-30
api: custom packer and unpacker factories

DifferentialOrange:
After this patch, user may pass `packer_factory` and `unpacker_factory` options to a connection. They will be used instead of the default ones. `packer_factory` is expected to be a function with the only one parameter: connection object, which returns a new msgpack.Packer object. `unpacker_factory` is expected to be a function with the only one parameter: connection object, which returns a new msgpack.Unpacker object. `packer_factory` supersedes `encoding` option. `unpacker_factory` supersedes `encoding` and `use_list` options. User may implement `encoding` and `use_list` support in its custom packer or unpacker if they wish so. User may refer to request submodule `packer_factory` and response submodule `unpacker_factory` as an example (these factories are used by default.)

Closes #145, #190, #191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Support custom packer and unpacker factories (#191).
 
 ### Changed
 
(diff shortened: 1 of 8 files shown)


GRISHNOV (approved):
Hi! LGTM

PR #271, opened 2022-12-12
api: extend connect with `fetch_schema` param

### LINKED ISSUE #219: Allow to disable schema fetch
I suggest to implement an option that disables schema fetching.

As it was implemented in netbox (https://github.com/tarantool/tarantool/issues/4789).
And as it was implemented in Go connector:
https://github.com/tarantool/go-tarantool/blob/de95e3184df6bde148e7cc87229572d4247f599b/connection.go#L199-L201

GRISHNOV:
Added support of the `fetch_schema` parameter, which allows to ignore schema changes on the server.

By default, it is used `fetch_schema = True`:

```python
>>> import tarantool
>>> conn = tarantool.Connection(host='localhost', port=3303)

>>> conn.schema_version
80

>>> conn.schema.schema
{257: <tarantool.schema.SchemaSpace object at 0x10fd21a80>, '_vinyl_deferred_delete': <tarantool.schema.SchemaSpace object at 0x10fd21a80>, 272: <tarantool.schema.SchemaSpace object at 0x10fd229b0>, ... ... , 'tester': <tarantool.schema.SchemaSpace object at 0x10fd22830>}

>>> tester = conn.schema.schema['tester']
>>> tester
<tarantool.schema.SchemaSpace object at 0x10fd22830>
>>> tester.
tester.arity    tester.flush()  tester.format   tester.indexes  tester.name     tester.schema   tester.sid      
>>> tester.format
{'id': {'name': 'id', 'type': 'unsigned', 'id': 0}, 0: {'name': 'id', 'type': 'unsigned', 'id': 0}, 'name': {'name': 'name', 'type': 'string', 'id': 1}, 1: {'name': 'name', 'type': 'string', 'id': 1}}
```

If the `fetch_schema` is specified as `False`, fields `schema_version` and `schema` will no longer be present in the `connection` object:

```python
>>> import tarantool
>>> conn = tarantool.Connection(host='localhost', port=3303, fetch_schema=False)

>>> hasattr(conn, 'schema_version')
False
>>> hasattr(conn, 'schema')
False
```

In this case, requests to the server via the `connection.update_schema` will no longer be made when there is `SchemaReloadException`.

Closes #219

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Support `fetch_schema` parameter for a connection (#219).
 
 ### Added
 
--- a/tarantool/connection.py
+++ b/tarantool/connection.py
@@ -579,7 +579,8 @@ class Connection(ConnectionInterface):
                  ssl_password_file=DEFAULT_SSL_PASSWORD_FILE,
                  packer_factory=default_packer_factory,
                  unpacker_factory=default_unpacker_factory,
-                 auth_type=None):
+                 auth_type=None,
+                 fetch_schema=True):
         """
         :param host: Server hostname or IP address. Use ``None`` for
             Unix sockets.
@@ -736,6 +737,18 @@ class Connection(ConnectionInterface):
             ``"chap-sha1"``.
         :type auth_type: :obj:`None` or :obj:`str`, optional
 
+        :param bool fetch_schema: If ``False``, schema is not loaded on connect
+            and schema updates are not automatically loaded.
+            As a result, these methods become unavailable:
+            :meth:`~tarantool.Connection.replace`,
+            :meth:`~tarantool.Connection.insert`,
+            :meth:`~tarantool.Connection.delete`,
+            :meth:`~tarantool.Connection.upsert`,
+            :meth:`~tarantool.Connection.update`,
+            :meth:`~tarantool.Connection.select`,
+            :meth:`~tarantool.Connection.space`.
+        :type fetch_schema: :obj:`bool`, optional
+
         :raise: :exc:`~tarantool.error.ConfigurationError`,
             :meth:`~tarantool.Connection.connect` exceptions
 
@@ -766,8 +779,9 @@ class Connection(ConnectionInterface):
         self.socket_timeout = socket_timeout
         self.reconnect_delay = reconnect_delay
         self.reconnect_max_attempts = reconnect_max_attempts
-        self.schema = Schema(self)
-        self.schema_version = 1
+        self.fetch_schema = fetch_schema
+        self.schema = None
+        self.schema_version = 0
         self._socket = None
         self.connected = False
         self.error = True
@@ -1023,7 +1037,11 @@ class Connection(ConnectionInterface):
             if self.transport == SSL_TRANSPORT:
                 self.wrap_socket_ssl()
             self.handshake()
-            self.load_schema()
+            if self.fetch_schema:
+                self.schema = Schema(self)
+                self.load_schema()
+            else:
+                self.schema = None
         except SslError as e:
             raise e
         except Exception as e:
@@ -1118,7 +1136,8 @@ class Connection(ConnectionInterface):
                 response = request.response_class(self, self._read_response())
                 break
             except SchemaReloadException as e:
-                self.update_schema(e.schema_version)
+                if self.schema is not None:
+                    self.update_schema(e.schema_version)
                 continue
 
         while response._code == IPROTO_CHUNK:
@@ -1255,6 +1274,9 @@ class Connection(ConnectionInterface):
         :meta private:
         """
 
+        if self.schema is None:
+            self.schema = Schema(self)
+
         self.schema_version = schema_version
         self.flush_schema()
 
@@ -1268,6 +1290,19 @@ class Connection(ConnectionInterface):
 
         self.schema.flush()
         self.load_schema()
+
+    def _schemaful_connection_check(self):
+        """
+        Checks whether the connection is schemaful.
+        If the connection is schemaless, an exception will be thrown
+        about unsupporting the method in connection opened
+        with fetch_schema=False.
+
+        :raise: :exc:`~tarantool.error.NotSupportedError`
+        """
+        if self.schema is None:
+            raise NotSupportedError('This method is not available in ' +
+                                    'connection opened with fetch_schema=False')
 
     def call(self, func_name, *args, on_push=None, on_push_ctx=None):
         """
@@ -1366,10 +1401,13 @@ class Connection(ConnectionInterface):
             :exc:`~tarantool.error.DatabaseError`,
             :exc:`~tarantool.error.SchemaError`,
             :exc:`~tarantool.error.NetworkError`,
-            :exc:`~tarantool.error.SslError`
+            :exc:`~tarantool.error.SslError`,
+            :exc:`~tarantool.error.NotSupportedError`
 
         .. _replace: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/replace/
         """
+
+        self._schemaful_connection_check()
 
         if isinstance(space_name, str):
             space_name = self.schema.get_space(space_name).sid
@@ -1411,7 +1449,7 @@ class Connection(ConnectionInterface):
                               password=self.password,
                               auth_type=self._get_auth_type())
         auth_response = self._send_request_wo_reconnect(request)
-        if auth_response.return_code == 0:
+        if auth_response.return_code == 0 and self.schema is not None:
             self.flush_schema()
         return auth_response
 
@@ -1584,10 +1622,13 @@ class Connection(ConnectionInterface):
             :exc:`~tarantool.error.DatabaseError`,
             :exc:`~tarantool.error.SchemaError`,
             :exc:`~tarantool.error.NetworkError`,
-            :exc:`~tarantool.error.SslError`
+            :exc:`~tarantool.error.SslError`,
+            :exc:`~tarantool.error.NotSupportedError`
 
         .. _insert: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/insert/
         """
+
+        self._schemaful_connection_check()
 
         if isinstance(space_name, str):
             space_name = self.schema.get_space(space_name).sid
@@ -1623,10 +1664,13 @@ class Connection(ConnectionInterface):
             :exc:`~tarantool.error.DatabaseError`,
             :exc:`~tarantool.error.SchemaError`,
             :exc:`~tarantool.error.NetworkError`,
-            :exc:`~tarantool.error.SslError`
+            :exc:`~tarantool.error.SslError`,
+            :exc:`~tarantool.error.NotSupportedError`
 
         .. _delete: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/delete/
         """
+
+        self._schemaful_connection_check()
 
         key = wrap_key(key)
         if isinstance(space_name, str):
@@ -1682,10 +1726,13 @@ class Connection(ConnectionInterface):
             :exc:`~tarantool.error.DatabaseError`,
             :exc:`~tarantool.error.SchemaError`,
             :exc:`~tarantool.error.NetworkError`,
-            :exc:`~tarantool.error.SslError`
+            :exc:`~tarantool.error.SslError`,
+            :exc:`~tarantool.error.NotSupportedError`
 
         .. _upsert: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/upsert/
         """
+
+        self._schemaful_connection_check()
 
         if isinstance(space_name, str):
             space_name = self.schema.get_space(space_name).sid
@@ -1770,10 +1817,13 @@ class Connection(ConnectionInterface):
             :exc:`~tarantool.error.DatabaseError`,
             :exc:`~tarantool.error.SchemaError`,
             :exc:`~tarantool.error.NetworkError`,
-            :exc:`~tarantool.error.SslError`
+            :exc:`~tarantool.error.SslError`,
+            :exc:`~tarantool.error.NotSupportedError`
 
         .. _update: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/update/
         """
+
+        self._schemaful_connection_check()
 
         key = wrap_key(key)
         if isinstance(space_name, str):
@@ -1956,10 +2006,13 @@ class Connection(ConnectionInterface):
             :exc:`~tarantool.error.DatabaseError`,
             :exc:`~tarantool.error.SchemaError`,
             :exc:`~tarantool.error.NetworkError`,
-            :exc:`~tarantool.error.SslError`
+            :exc:`~tarantool.error.SslError`,
+            :exc:`~tarantool.error.NotSupportedError`
 
         .. _select: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/select/
         """
+
+        self._schemaful_connection_check()
 
         if iterator is None:
             iterator = ITERATOR_EQ
@@ -1995,6 +2048,8 @@ class Connection(ConnectionInterface):
 
         :raise: :exc:`~tarantool.error.SchemaError`
         """
+
+        self._schemaful_connection_check()
 
         return Space(self, space_name)
 
--- a/test/suites/crud_server.lua
+++ b/test/suites/crud_server.lua
@@ -1,7 +1,58 @@
 #!/usr/bin/env tarantool
 
-local crud = require('crud')
-local vshard = require('vshard')
+local function configure_crud_instance(primary_listen, crud, vshard)
+    box.schema.create_space(
+    'tester', {
+    format = {
+        {name = 'id', type = 'unsigned'},
+        {name = 'bucket_id', type = 'unsigned'},
+        {name = 'name', type = 'string'},
+    }
+    })
+    box.space.tester:create_index('primary_index', {
+        parts = {
+            {field = 1, type = 'unsigned'},
+        },
+    })
+    box.space.tester:create_index('bucket_id', {
+        parts = {
+            {field = 2, type = 'unsigned'},
+        },
+        unique = false,
+    })
+
+    -- Setup vshard.
+    _G.vshard = vshard
+    box.once('guest', function()
+        box.schema.user.grant('guest', 'super')
+    end)
+    local uri = 'guest@0.0.0.0:' .. primary_listen
+    local cfg = {
+        bucket_count = 300,
+        sharding = {
+            [box.info().cluster.uuid] = {
+                replicas = {
+                    [box.info().uuid] = {
+                        uri = uri,
+                        name = 'storage',
+                        master = true,
+                    },
+                },
+            },
+        },
+    }
+    vshard.storage.cfg(cfg, box.info().uuid)
+    vshard.router.cfg(cfg)
+    vshard.router.bootstrap()
+
+    -- Initialize crud.
+    crud.init_storage()
+    crud.init_router()
+    crud.cfg{stats = true}
+end
+
+local crud_imported, crud = pcall(require, 'crud')
+local vshard_imported, vshard = pcall(require, 'vshard')
 
 local admin_listen = os.getenv("ADMIN")
 local primary_listen = os.getenv("LISTEN")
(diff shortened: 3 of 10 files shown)


GRISHNOV:
> > In this case, requests to the server via the connection.update_schema will no longer be made when there is SchemaReloadException.
> 
> So there won't be any way to use a connection for any space operations? Only calls and evals? To be honest, I don't quite get this one without tests.

I suppose, it is.
By analogy with an existing solution:
```
c = netbox.connect(uri, {fetch_schema = false})
c.space -- always will be nil
```

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Support custom packer and unpacker factories (#191).
```
DifferentialOrange, line 11: We don't add newlines between changelog entries.

tarantool/connection.py
```diff
@@ -503,6 +504,17 @@ def __init__(self, host, port,
...
 
+        :param bool fetch_schema: If ``False``, allows to ignore schema
+            changes on the server, schema updates are not automatically
+            loaded. As a result, the methods become unavailable:
+            :meth:`~tarantool.Connection.replace`,
+            :meth:`~tarantool.Connection.insert`,
+            :meth:`~tarantool.Connection.delete`,
+            :meth:`~tarantool.Connection.upsert`,
+            :meth:`~tarantool.Connection.update`,
+            :meth:`~tarantool.Connection.select`.
```
DifferentialOrange, line 515: AFAIU, `space` is also unavailable

tarantool/connection.py
```diff
@@ -533,8 +545,10 @@ def __init__(self, host, port,
         self.socket_timeout = socket_timeout
         self.reconnect_delay = reconnect_delay
         self.reconnect_max_attempts = reconnect_max_attempts
-        self.schema = Schema(self)
-        self.schema_version = 1
+        self.fetch_schema = fetch_schema
+        if self.fetch_schema:
+            self.schema = Schema(self)
```
DifferentialOrange, line 550: Let's set them to `None` otherwise

test/suites/test_schema.py
```diff
@@ -43,6 +44,8 @@ def setUpClass(self):
         self.srv.start()
         self.con = tarantool.Connection(self.srv.host, self.srv.args['primary'],
                                         encoding=self.encoding)
+        self.con_schema_disable = tarantool.Connection(self.srv.host, self.srv.args['primary'],
+                                                       encoding=self.encoding, fetch_schema=False)
```
DifferentialOrange, line 48: `authenticate` reloads schema. For now, it will fail on attempt to subscribe `self.schema` in `load_schema`. So your tests have two major issues (related to each other):
- There is no authenticate. `self.con` authenticates in `test_00` -- it is a very weird and misleading approach to use separate test case to prepare something and use `test_N` naming to rely on their tests. I think we may simply add `user`/`password` here.
- There is no check that every other handle is ok. For example, because of missing authenticate, `eval` throws `tarantool.error.DatabaseError: (42, "Execute access to universe '' is denied for user 'guest'")`.

We must fix both of them.

tarantool/connection.py
```diff
@@ -503,6 +504,17 @@ def __init__(self, host, port,
             callable[[:obj:`~tarantool.Connection`], :obj:`~msgpack.Unpacker`],
             optional
 
+        :param bool fetch_schema: If ``False``, allows to ignore schema
+            changes on the server, schema updates are not automatically
+            loaded. As a result, the methods become unavailable:
```
DifferentialOrange, line 509:
```suggestion
            loaded. As a result, these methods become unavailable:
```

tarantool/connection.py
```diff
@@ -363,7 +363,8 @@ def __init__(self, host, port,
                  ssl_ca_file=DEFAULT_SSL_CA_FILE,
                  ssl_ciphers=DEFAULT_SSL_CIPHERS,
                  packer_factory=default_packer_factory,
-                 unpacker_factory=default_unpacker_factory):
+                 unpacker_factory=default_unpacker_factory,
+                 fetch_schema=True):
```
DifferentialOrange, line 367: Since `ConnectionPool` does not inherit `Connection` (I had reasons for that), you need to add new parameter explicitly.

tarantool/connection.py
```diff
@@ -1089,11 +1105,16 @@ def replace(self, space_name, values, on_push=None, on_push_ctx=None):
...
             :exc:`~tarantool.error.SchemaError`,
             :exc:`~tarantool.error.NetworkError`,
-            :exc:`~tarantool.error.SslError`
+            :exc:`~tarantool.error.SslError`,
+            :exc:`~tarantool.error.NotSupportedError`
 
         .. _replace: https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_space/replace/
         """
 
+        if not self.fetch_schema:
```
DifferentialOrange, line 1114: Well, since all asserts are the same, I think we may extract them to some function.

DifferentialOrange, line 1114: Btw, even though `load_schema` marked `:meta private:` in documentation, I think it's possible for one to load schema some time after even if they set `self.fetch_schema` to `false`. So
- let's check not the `fetch_schema` option, but the presence of `schema` itself,
- let's add a test that it's possible to call `load_schema`/`flush_schema` and make our connection schemafull again.

DifferentialOrange (changes requested):
Thank you for your PR and sorry I'm a bit late (I wasn't sure that it was ready for review).

tarantool/connection.py
DifferentialOrange, line 367: Let's also use `fetch_schema=False` connection in crud API example.

test/suites/test_schema.py
DifferentialOrange: Let's also test it        like crud: for all three types of connection
https://github.com/tarantool/tarantool-python/blob/6f0d5369ab115c8e7d456e8b43ae77d80f5de705/test/suites/test_crud.py#L711

DifferentialOrange:
You'll also need to rebase, sorry for the inconveniences

CHANGELOG.md
GRISHNOV: Done

tarantool/connection.py

tarantool/connection.py
GRISHNOV: Related with [this](https://github.com/tarantool/tarantool-python/pull/271#discussion_r1056184340).
If I understand correctly, to be able to make the connection again schema-full without reconnect, `None` is not needed here

test/suites/test_schema.py
GRISHNOV: Done
* Test case `test_00` has been deleted. Authentication is performed when opening connections to test instances
* Now not only the unavailability of some methods is checked, but also the availability of those that should not have been affected
* Implemented with `connection`, `mesh_connection` and `pool_connection`

tarantool/connection.py
GRISHNOV: Fixed

tarantool/connection.py
GRISHNOV: Implemented

tarantool/connection.py
GRISHNOV: Done, implemented `_fetch_schema_support_check`

GRISHNOV: Done, now the connection `fetch_schema` type can be changed. Also added tests with this case

tarantool/connection.py
GRISHNOV: Done. Also now `fetch_schema=False` used in CRUD tests

GRISHNOV:
While working on the task, I noticed that there is an issue in the CRUD tests.
When running `make test` locally, in the absence of an installed `crud rocks`, CRUD tests freezes forever. 
So, I made some changes to the CRUD instance configuration file

test/suites/crud_server.lua
```diff
@@ -18,51 +69,19 @@ box.schema.user.grant(
...
+                     'installation via rocks install is required ' ..
+                     'for CRUD testing purposes. You can use ' ..
+                     '<tarantoolctl rocks install crud> or ' ..
+                     '<tt rocks install crud> to install modules'
+    -- The print output will be captured in the logs.
+    print(fail_msg)
+    -- Because of the implementation of the testing framework,
+    -- this will cause the crash of the CRUD test and display to user
+    -- a message about require fail during instance configuration.
+    box.info.status = fail_msg
```
DifferentialOrange, line 84: This hack is definitely not nice. Let's do something other here, if possible

test/suites/crud_server.lua
```diff
@@ -18,51 +69,19 @@ box.schema.user.grant(
...
-}
-vshard.storage.cfg(cfg, box.info().uuid)
-vshard.router.cfg(cfg)
-vshard.router.bootstrap()
 
--- Initialize crud.
-crud.init_storage()
-crud.init_router()
-crud.cfg{stats = true}
+if crud_imported == false or vshard_imported == false then
```
DifferentialOrange, line 73: And if you want to fix crud tests here, it should be a separate commit.

test/suites/test_schema.py
```diff
@@ -97,13 +135,7 @@ def verify_unicode_index(self, index):
...
         self.assertEqual(len(index.parts), 1)
 
-    def test_00_authenticate(self):
-        self.assertIsNone(self.srv.admin("box.schema.user.create('test', { password = 'test' })"))
-        self.assertIsNone(self.srv.admin("box.schema.user.grant('test', 'read,write', 'space', '_space')"))
-        self.assertIsNone(self.srv.admin("box.schema.user.grant('test', 'read,write', 'space', '_index')"))
-        self.assertEqual(self.con.authenticate('test', 'test')._data, None)
-
-    def test_01_space_bad(self):
+    def test_00_space_bad(self):
```
DifferentialOrange, line 138: You can just leave it here starting from 01

test/suites/test_schema.py
```diff
@@ -336,14 +368,163 @@ def test_06_index_cached(self):
...
+            }, 
+            'select': {
+                'input': ['tester', 1],
+                'output': [[1, None]],
+            },
+            'space': {
+                'input': ['tester'],
+            },
+        },
+        'available': {
```
DifferentialOrange, line 407: ping?

DifferentialOrange, line 407: Leave a comment here referring to `test_crud.py` regarding crud methods.

DifferentialOrange, line 407: SQL?

tarantool/connection.py
DifferentialOrange: No, it shouldn't be like this. Changing internal structure parameter is not a good approach. I think you should simply
- do not fetch schema on connect if you wasn't asked to do so,
- do not allow to run requests using schema if there is not any schema,
- do not reload schema is there is none,
- allow to run requests using schema if there is one,
- reload schema is there is one.

Maybe there are some non-obvious points regarding schema clean-up between reconnects.

DifferentialOrange (changes requested):
Thank you for the updates! There are still some things we need to resolve

tarantool/connection.py
```diff
@@ -1023,7 +1037,9 @@ def connect(self):
             if self.transport == SSL_TRANSPORT:
                 self.wrap_socket_ssl()
             self.handshake()
-            self.load_schema()
+            if self.fetch_schema:
+                self.load_schema()
+            self._check_features()
```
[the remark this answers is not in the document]
GRISHNOV: Done, deleted `self._check_features()`

test/suites/crud_server.lua
GRISHNOV: Unfortunately, I don't see any other options yet. The [design](https://github.com/tarantool/tarantool-python/blob/master/test/suites/lib/tarantool_server.py#L237) of the test framework, as far as I can see, does not allow you to get a more accurate result without test framework changes. It may make sense to implement exit conditions from waiting in `wait_until_started`. I'll try to consider this

test/suites/crud_server.lua
GRISHNOV: Done with separate commit

test/suites/test_schema.py
GRISHNOV: Done

test/suites/test_schema.py
GRISHNOV: Added

GRISHNOV: Done

GRISHNOV: If I understand correctly, this is about the `execute` method. It is added in the method `test_08_schema_fetch_disable` (depending on tarantool version)

test/suites/crud_server.lua
DifferentialOrange: Well, if you don't find anything better in reasonable time (not more than a couple of hours of work), let's leave it as is for now.

DifferentialOrange: But I'm not sure this will work fine since `box.info.status` is highly likely to get overwritten by Tarantool

GRISHNOV: > Let's do something relatively easy here, like pcalling `crud` in init file and checking in `setUp` if require is ok and failing (or skipping) otherwise

Done, if the solution is suitable, I will make the correct division into commits

DifferentialOrange: Yeah, I think it's suitable, thank you.

tarantool/connection.py
GRISHNOV: Do I understand correctly:

* The `fetch_schema` attribute in the connection object should not be available to the user, it should be made private (now it is possible to change the parameter like `conn.fetch_schema=...` for an existing `conn` object, now `fetch_schema` is public). We want the `fetch_schema` parameter is manually set by user only when establishing a connection.
    + If it is decided to make the `fetch_schema` private, then we can implement a getter (`conn.is_fetch_schema_enable() -> Bool`), so we can understand the type of connection.
* However, the connection can be made from `schema-less` to `schema-full`. To do this, it is enough for the user to call `conn.update_schema(conn.schema_version)`. After that, all methods that were previously unavailable for `schema-less` connection become available. At the same time, the scheme starts updating automatically. And `conn.is_fetch_schema_enable()` will start returning `True`.
* It is not possible to change the connection type from `schema-full` to `schema-less`.

DifferentialOrange: > The fetch_schema attribute in the connection object should not be available to the user, it should be made private (now it is possible to change the parameter like conn.fetch_schema=... for an existing conn object, now fetch_schema is public)

There are no convenient in-built ways to declare a "private" in Python class: either you start it with underscore (and it still would be public) or mess with getters/setters overwriting, similar to Lua (and it's mostly don't worth it). You can leave it as is now, but never rely on someone would change that and do not teach others to do so through examples and tests.

> * If it is decided to make the `fetch_schema` private, then we can implement a getter (`conn.is_fetch_schema_enable() -> Bool`), so we can understand the type of connection.

It's also not worth it. I don't think that it be used that much often, and if one really needs it, he can just look up our "private" variable.



> * However, the connection can be made from `schema-less` to `schema-full`. To do this, it is enough for the user to call `conn.update_schema(conn.schema_version)`. After that, all methods that were previously unavailable for `schema-less` connection become available. At the same time, the scheme starts updating automatically. And `conn.is_fetch_schema_enable()` will start returning `True`.

> * It is not possible to change the connection type from `schema-full` to `schema-less`.

Yeah, I propose to use this approach.

GRISHNOV: AFAIU, done. Now switching from `schema-less` to `schema-full` possible via `conn.update_schema(conn.schema_version)`, tests also cover this case

DifferentialOrange: Why we still base our conditions on parameter value rather that schema value itself?

DifferentialOrange: I don't think we should ever overwrite `fetch_schema` parameter which should be used to determine if schema would be fetched on connect or not and only there.

> Why we still base our conditional checks on parameter value rather that schema value itself?

Now the presence of the schema value is checked

> I don't think we should ever overwrite `fetch_schema` parameter which should be used to determine if schema would be fetched on connect or not and only there.

Now this value changes only inside `update_schema` from `False` to `True`

tarantool/connection.py
```diff
@@ -1023,7 +1040,8 @@ def connect(self):
             if self.transport == SSL_TRANSPORT:
                 self.wrap_socket_ssl()
             self.handshake()
-            self.load_schema()
+            if self.schema.schema != {}:
```
DifferentialOrange, line 1043: Relying on internals in such a way is not the best approach. And I think this one is also wrong -- `schema.schema` is `{}` is there are no spaces yet.

tarantool/connection.py
```diff
@@ -1023,7 +1040,8 @@ def connect(self):
             if self.transport == SSL_TRANSPORT:
                 self.wrap_socket_ssl()
             self.handshake()
-            self.load_schema()
```
DifferentialOrange, line 1026: It seems that my request to support explicit `update_schema` call is not so simple to implement (and it doesn't seem like a good design since its documented as a private API). Let's forget this one (since such use case is rather weird) and simply implement the following:
- initialize `self.schema` to `None`;
- if `fetch_schema` is `True`, build initial `self.schema` here, on `connect`, and load it;
- if `fetch_schema` is `False`, set `self.schema` to `None` here, on `connect` (just in case);
- if `self.schema` is `None`, raise errors on requests with schema;
- if `self.schema` is not `None`, reload the schema on receiving `SchemaReloadRequest`.

With this approach, `fetch_schema` meaning is "whether we fetch schema on connect or not", which is, I think, rather straightforward.

Let's remove explicit `update_schema`/`load_schema` examples from test. Since Python allows to mess with object internals, it still would be possible to turn schemaless connection to schemaful, but it shouldn't be our burden anymore.

DifferentialOrange: Maybe schema_version reset should also be processed

tarantool/connection.py
```diff
@@ -1118,7 +1136,8 @@ def _send_request_wo_reconnect(self, request, on_push=None, on_push_ctx=None):
                 response = request.response_class(self, self._read_response())
                 break
             except SchemaReloadException as e:
-                self.update_schema(e.schema_version)
+                if self.schema.schema != {}:
```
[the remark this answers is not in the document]
GRISHNOV: [Done](https://github.com/tarantool/tarantool-python/pull/271#discussion_r1065453318)

tarantool/connection.py
```diff
@@ -736,6 +737,21 @@ def __init__(self, host, port,
...
+            loaded. As a result, these methods become unavailable:
+            :meth:`~tarantool.Connection.replace`,
+            :meth:`~tarantool.Connection.insert`,
+            :meth:`~tarantool.Connection.delete`,
+            :meth:`~tarantool.Connection.upsert`,
+            :meth:`~tarantool.Connection.update`,
+            :meth:`~tarantool.Connection.select`,
+            :meth:`~tarantool.Connection.space`.
+            Calling the :meth:`~tarantool.Connection.update_schema`
+            method causes this parameter to switch
```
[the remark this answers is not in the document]
GRISHNOV: Done, deleted

tarantool/connection.py
```diff
@@ -1269,6 +1291,18 @@ def flush_schema(self):
...
 
+    def _fetch_schema_support_check(self):
+        """
+        Checks the fetch_schema opt of the connection is enabled.
+        If the opt is disabled, an exception will be thrown 
+        about unsupporting the method with fetch_schema=False.
+
+        :raise: :exc:`~tarantool.error.NotSupportedError`
+        """
+        if not self.fetch_schema:
```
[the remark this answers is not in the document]
GRISHNOV: Done, now `if self.schema is None:`

tarantool/connection.py
GRISHNOV: [Done](https://github.com/tarantool/tarantool-python/pull/271#discussion_r1065453318)

tarantool/connection.py
GRISHNOV: Done

> Let's remove explicit update_schema/load_schema examples from test. Since Python allows to mess with object internals, it still would be possible to turn schemaless connection to schemaful, but it shouldn't be our burden anymore.

I don't quite understand what is meant, now `update_schema` is only used inside `test_08_schema_fetch_disable` when switching from schemaless to schemaful

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to decrypt private SSL key file (#224).
 - Support specifying authentication method with `auth_type`
   and Tarantool EE `pap-sha256` authentication method (#269).
+- Support of the fetch_schema parameter for connection (#219).
```
DifferentialOrange, line 24:
```suggestion
- Support `fetch_schema` parameter for a connection (#219).
```

tarantool/connection.py
```diff
@@ -1269,6 +1291,18 @@ def flush_schema(self):
         self.schema.flush()
         self.load_schema()
 
+    def _fetch_schema_support_check(self):
+        """
+        Checks the fetch_schema opt of the connection is enabled.
```
DifferentialOrange, line 1296: Description is outdated

DifferentialOrange (approved):
Thank you for the updates! Let's fix remaining minor things and merge this one

GRISHNOV:
Thanks for the feedback! 

I've made all the corrections. At the moment there is a problem with tests that fall with variable probability. At the moment, it is difficult to determine the cause, but the problem, as far as I can see, is related to the `update_schema` call for connection pool:

```python
# Turning the same connection into schemafull.
if mode is not None:
    for addr in con.pool.keys():
        con.pool[addr].conn.update_schema(con.pool[addr].conn.schema_version) # <------
    else:
        con.update_schema(con.schema_version)
```

It looks like there is a desynchronization with the transmitted `con.pool[addr].conn.schema_version` value and the value on the server

CHANGELOG.md
GRISHNOV: Done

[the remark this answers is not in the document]
GRISHNOV: Done

tarantool/connection.py
GRISHNOV: Description updated and func renamed to `_schemafull_connection_check`

tarantool/connection.py
```diff
@@ -736,6 +737,18 @@ def __init__(self, host, port,
             ``"chap-sha1"``.
         :type auth_type: :obj:`None` or :obj:`str`, optional
 
+        :param bool fetch_schema: If ``False``, allows to ignore schema
```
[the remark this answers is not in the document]

test/suites/test_schema.py
```diff
@@ -341,9 +372,169 @@ def test_07_schema_version_update(self):
...
+                        resp = testing_function(
+                            *self.testing_methods['available'][method_case]['input'])
+                    if method_case == 'ping':
+                        self.assertEqual(isinstance(resp, float), True)
+                    else:
+                        self.assertEqual(
+                            resp.data,
+                            self.testing_methods['available'][method_case]['output'])
+
+            # Turning the same connection into fetch_schema=True.
```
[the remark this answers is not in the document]
GRISHNOV: Updated

test/suites/test_schema.py
```diff
@@ -341,9 +372,169 @@ def test_07_schema_version_update(self):
...
+            if mode is not None:
+                for addr in con.pool.keys():
+                    con.pool[addr].conn.update_schema(con.pool[addr].conn.schema_version)
+            else:
+                con.update_schema(con.schema_version)
+
+            # Testing the fetch_schema=True connection with methods
```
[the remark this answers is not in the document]
GRISHNOV: Updated

tarantool/connection.py
```diff
@@ -1269,6 +1291,19 @@ def flush_schema(self):
         self.schema.flush()
         self.load_schema()
 
+    def _schemafull_connection_check(self):
```
DifferentialOrange, line 1294: It seems that `schemaful` is a more valid word (even though they both are not in dictionaries)
https://github.com/surrealdb/surrealdb/issues/115

DifferentialOrange (approved):
Everything seems good, I'll wait till you resolve this flaky test issue.

PR #272, opened 2022-12-22
Fix packing in Github Actions

DifferentialOrange:
### ci: use testing rules in packing

Use the same rules in packing as in testing to run on dev branches.

Follows #248

### ci: fix crud install

Set valid package manager for fedora. Fix using apt for docker:
debian/ubuntu containers are sudoless.

Follows #264

### ci: fixate OS versions

deb packages built with Ubuntu 22.04 fails to install on Debian 10, 11.

### DIFF
--- a/.github/workflows/packing.yml
+++ b/.github/workflows/packing.yml
@@ -2,10 +2,6 @@ name: packing
 
 on:
   push:
-    branches:
-      - master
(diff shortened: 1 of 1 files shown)


GRISHNOV (approved):
Hi! LGTM

Issue #224, opened 2022-08-10
Support SSL passwords


DifferentialOrange:
`ssl_password` and `ssl_password_file` were introduced in Tarantool EE (yet unreleased): https://github.com/tarantool/tarantool-ee/commit/e1f47dd4adbc6657159c611298aad225883a536b .

PR #273, opened 2022-12-23
api: support authentication methods

### LINKED ISSUE #269: Support new authentication mechanisms
New authentication mechanisms are in design for next Tarantool versions. After the release of this feature, we must support it.

DifferentialOrange:
Merge after #274

Since Tarantool master 2574ff1a configuring authentication type is
supported [1-2]. Together with this, Tarantool EE had introduced
pap-sha256 authentication method support [3-4]. It can be used only
together with SSL transport. To configure, use `auth_type` option in
Connection, MeshConnection or ConnectionPool.

Newest master (there is no dev build in customer zone yet) supports
providing authentication method via IPROTO_ID response. So in this patch
we also move ID request to be executed before authentication.

1. https://github.com/tarantool/tarantool/issues/7988
2. https://github.com/tarantool/tarantool/issues/7989
3. https://github.com/tarantool/tarantool-ee/issues/295
4. https://github.com/tarantool/tarantool-ee/issues/322

Closes #269


oleg-jukovec (approved):
Thank you for the patch!

DifferentialOrange:
> Thank you for the patch!

Thank you for the review! Can you also check this one https://github.com/tarantool/tarantool-python/issues/224 ?

PR #277, opened 2023-01-31
fix: Misspell in return error code

bekhzod91:


DifferentialOrange (commented):
Thank you for your PR! Can you add a changelog entry? And, please, use lower case in commit message ("Misspell" to "misspell").

It's a shame that we still haven't got a linter to prevent this, but there are too many things we'll need to rework beforehand.

DifferentialOrange:
Some of the CI pipelines are red now, but this is definitely on my side.

PR #291, opened 2023-04-14
Get rid of pandas dependency

### LINKED ISSUE #290: Make pandas dependency optional
There are a lot of projects with Tarantool+Python which are not interested in using datetimes (for example, all projects with Tarantool 2.8 and older). They are still obliged to install `pandas` and `pytz`, both are rather heavy. Users ask to make this dependency optional.

It seems that Python package mechanism for providing optional dependencies is ["extras"](https://peps.python.org/pep-0440/). We also shouldn't forget to make deb and rpm package dependencies optional.

This change is a **breaking** one, since user would need to change their setup process if they want to work with datetimes.

We may start throwing errors for datetimes in msgpack if we cannot parse them or return some dictionaries/tuples with raw binary data.

DifferentialOrange:
Rework our implementation of tarantool.Datetime class. Previously it had relied on pandas.Timestamp and pandas.Timedelta. There were user complaints about pandas as a requirement since it's rather heavy. Now our implementation of datetime uses built-in datetime.datetime, datetime.timedelta and some other built-in tools.

It is expected that the implementation change wouldn't affect users, but some minor behavior traits were broken in this patch:
- Now we rely on datetime argument validation which differs from the pandas one. For example, it doesn't allow overflows for fields.  Exceptions that user may receive from internal datetime are, of course, had changed as well.
- We drop the support of `__eq__` for pandas.Timestamp. We simply compared underlying pandas.Timestamp with argument one, and now it's impossible. If the feature would be useful later, we may implement the comparison in some compatible way.
- `__repr__` has been changed since internal representation has been changed as well.

Closes #290

### DIFF
--- a/tarantool/msgpack_ext/types/datetime.py
+++ b/tarantool/msgpack_ext/types/datetime.py
@@ -3,9 +3,11 @@ Tarantool `datetime`_ extension type implementation module.
 """
 # pylint: disable=line-too-long
 
+from calendar import monthrange
 from copy import deepcopy
+from datetime import datetime, timedelta
+import sys
 
-import pandas
 import pytz
 
 import tarantool.msgpack_ext.types.timezones as tt_timezones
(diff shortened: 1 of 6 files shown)


tarantool/msgpack_ext/types/datetime.py
```diff
@@ -495,23 +569,52 @@ def __eq__(self, other):
...
         if isinstance(other, Datetime):
-            return self._datetime == other._datetime
-        if isinstance(other, pandas.Timestamp):
-            return self._datetime == other
+            return self.value == other.value
         return False
 
-    def __str__(self):
-        return self._datetime.__str__()
+    def isoformat(self):
```
oleg-jukovec, line 581: Why do we need the method in the public API?

oleg-jukovec: I think it's better to hide it in a private method/function until anyone asks for it.
Up to you.

DifferentialOrange: Removed `isoformat`, moved code to `__str__` itself

PR #292, opened 2023-04-17
Remove join and subscribe & Release 1.0.0

DifferentialOrange:
## api: remove join and subscribe

This is a breaking change.

Current join and subscribe implementations are rather useless. Connector does not provide any API to process incoming replication requests. The only supported scenario is to "connect as replica, skip everything that has been sent through replication, close on error".

Current Tarantool team product strategy is to develop CDC features, including replication support in language connectors, as Enterprise edition products [1]. Since we don't plan to provide proper join and subscribe implementation in the open-source connector in the future, we decide to drop current half-baked implementation to not confuse new users.

1. https://github.com/tarantool/go-tarantool/issues/203

### Overview

This release introduces several minor behavior changes
to make API more consistent.

Starting from this release, connector no longer depends on `pandas`.

### Breaking changes

- Allow only named `on_push` and `on_push_ctx` for `insert` and `replace`.
- `tarantool.Datetime` `__repr__` has been changed.
- `tarantool.Datetime` input arguments are validated with `datetime.datetime` rules.
- `tarantool.Datetime` is no longer expected to throw `pandas.Timestamp`
  exceptions. `datetime.datetime` exceptions will be thrown instead of them.
- Drop the support of `__eq__` operator of `tarantool.Datetime` for `pandas.Timestamp`.
- Remove `join` and `subscribe` connection methods.

### Changes

- Migrate to built-in `Warning` instead of a custom one.
- Migrate to built-in `RecursionError` instead of a custom one.
- Collect full exception traceback.
- Package no longer depends on `pandas` (#290).

### Infrastructure

- Lint the code with `pylint`, `flake8` and `codespell`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     exceptions. `datetime.datetime` exceptions will
     be thrown instead of them.
   - Drop the support of `__eq__` operator for `pandas.Timestamp`.
+- **Breaking**: Remove `join` and `subscribe` connection methods.
 
 ## 0.12.1 - 2023-02-28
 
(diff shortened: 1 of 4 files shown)


LeonidVas (approved):
Hi! Thank you for the patchset.
LGTM!

PR #298, opened 2023-06-29
msgpack: support decimals with negative scale

DifferentialOrange:
Current decimal external type parser do not expect negative scale in decimal payload. Negative scale is a positive exponent. It seems that the only way to obtain positive exponent is to use E-notation since Tarantool library do not truncate trailing zeroes before decimal point:

```lua
tarantool> msgpack.encode(decimal.new('1e33')):hex()
---
- c70301d0df1c
...

tarantool> msgpack.encode(decimal.new('1000000000000000000000000000000000')):hex()
---
- c713010001000000000000000000000000000000000c
...
```

There are two different bugs in current implementation:
- we support only `positive fixint` scale and do not expect `int 8` [2],
- we do not expect negative scale so positive exponent will be ignored.

This patch fixes both of them. See also [3].

1. https://github.com/tarantool/tarantool/blob/ba749e820bf0638aa3f79f266848590f9713c1cf/src/lib/core/decimal.c#L432-L450
2. https://github.com/msgpack/msgpack/blob/master/spec.md
3. https://github.com/tarantool/go-tarantool/pull/314

### DIFF
--- a/test/suites/test_decimal.py
+++ b/test/suites/test_decimal.py
@@ -226,6 +226,39 @@ class TestSuiteDecimal(unittest.TestCase):
                         b'\x09\x87\x65\x43\x21\x98\x76\x54\x32\x1d'),
             'tarantool': "decimal.new('-1234567891234567890.0987654321987654321')",
         },
+        'decimal_exponent_1': {
+            'python': decimal.Decimal('1e33'),
+            'msgpack': (b'\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+                        b'\x00\x00\x00\x00\x00\x00\x00\x0c'),
+            'tarantool': "decimal.new('1e33')",
+        },
+        'decimal_exponent_2': {
+            'python': decimal.Decimal('1.2345e33'),
+            'msgpack': (b'\x00\x01\x23\x45\x00\x00\x00\x00\x00\x00\x00'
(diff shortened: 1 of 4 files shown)


test/suites/test_decimal.py
```diff
@@ -226,6 +226,39 @@ def setUp(self):
...
+        },
+        'decimal_exponent_5': {
+            'python': decimal.Decimal('-1e33'),
+            'msgpack': (b'\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00'
+                        b'\x00\x00\x00\x00\x00\x00\x00\x0d'),
+            'tarantool': "decimal.new('-1e33')",
+        },
+        'decimal_exponent_6': {
+            'python': decimal.Decimal('1e-33'),
+            'msgpack': (b'\x21\x1c'),
```
oleg-jukovec, line 259: Could you please add a comment: how to get the `msgpack` value for the test case?

oleg-jukovec (approved):
Thank you for the patch!

test/suites/test_decimal.py
DifferentialOrange: It's actually the same as in tarantool.

```
tarantool> msgpack.encode(decimal.new('1e-33')):hex()
---
- d501211c
...
```

PR #300, opened 2023-07-01
test: fix `box.tuple.new` variable argument syntax usage

CuriousGeorgiy:
Variable argument syntax of `box.tuple.new` will be disabled in scope of #4693: adapt the test suite to stop using it.


test/suites/test_dml.py
```diff
@@ -210,7 +210,7 @@ def test_07_call_16(self):
             self.assertEqual(len(ans[0]), 1)
             self.assertIsInstance(ans[0][0], str)
 
-            self.assertSequenceEqual(con.call('box.tuple.new', [1, 2, 3, 'fld_1']),
+            self.assertSequenceEqual(con.call('box.tuple.new', [[1, 2, 3, 'fld_1']]),
```
DifferentialOrange, line 213: Is `box.tuple.new('fid_1')` still valid? It is used in lines below.

PR #301, opened 2023-07-14
internal: add lost week in interval __str__

ArtDu:


DifferentialOrange:
Implemented as part of https://github.com/tarantool/tarantool-python/pull/302

PR #302, opened 2023-07-17
Fix intervals

DifferentialOrange:
Fix various issues related to datetime interval type. See commits for more info.

Thanks to @ArtDu for reporting these issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- `tarantool.Interval` arithmetic with weeks
+
 ## 1.1.0 - 2023-06-30
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - `tarantool.Interval` arithmetic with weeks
+- `tarantool.Interval` weeks display in `str()` and `repr()`
 
 ## 1.1.0 - 2023-06-30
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed
+- Validate `tarantool.Interval` limits with the same ruels as in Tarantool.
```
oleg-jukovec, line 10:
```suggestion
- Validate `tarantool.Interval` limits with the same rules as in Tarantool.
```

test/suites/test_interval.py
```diff
@@ -142,11 +174,263 @@ def test_interval_positional_init(self):
...
+                         r"datetime.interval.new({sec=1})",
+            'str': 'tarantool.Interval(year=0, month=0, week=0, day=0, hour=0, '
+                   f'minute=0, sec=-{int(MAX_SEC_RANGE)}, nsec=0, adjust=Adjust.NONE)',
+            'tarantool_8887_issue': True,
+        },
+        'max_sec_interval': {
+            'python': tarantool.Interval(sec=int(MAX_SEC_RANGE)),
+            'msgpack': (b'\x02\x06\xcf\x00\x01\x51\x81\xc6\x15\x59\x60\x08\x01'),
+            # Reason why Tarantool datetime is so weird here:
+            # https://github.com/tarantool/tarantool/issues/8878
```
oleg-jukovec, line 394: Will these tests break after the issue being fixed in Tarantool?

oleg-jukovec (approved):
Thank you for the patch!

test/suites/test_interval.py
oleg-jukovec: Could you please write it under a first commit with the weird behavior to make it clear?

```we will be able to simply write EXAMPLE after the fix```

CHANGELOG.md
DifferentialOrange: Thank you, added missing CHANGLELOG.md to codespell linting stage

test/suites/test_interval.py
DifferentialOrange: Added "Replace with X after fix" comments

PR #316, opened 2023-10-31
conn: create from socket fd

### LINKED ISSUE #304: Create connection from socket fd
Currently, to create a connection, the user has to specify a URI. The connector creates a socket and connects it to the given URI under the hood. There are use cases when the user doesn't have a URI but does have a connected socket fd. For example, a socket fd could be inherited from the parent process on fork. We need a way to wrap a raw socket fd number in a connection object.

For more details see the [IPROTO over socketpair](https://www.notion.so/tarantool/IPROTO-over-socketpair-fb5a30dc42a348689655bf75248d9020) design document.

askalt:
This patch adds the ability to create Tarantool connection using an existing socket fd.

I didn't forget about:
- [x] Tests
- [x] Changelog 

Closes #304

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -27,6 +27,7 @@ jobs:
           - '2.8'
           - '2.10'
           - '2.11'
+          - 'master'
         python:
           - '3.6'
           - '3.7'
@@ -57,10 +58,44 @@ jobs:
       - name: Clone the connector
         uses: actions/checkout@v3
 
+      - name: Setup tt
+        run: |
+          curl -L https://tarantool.io/release/2/installer.sh | sudo bash
+          sudo apt install -y tt
+          tt version
+          tt init 
+
       - name: Install tarantool ${{ matrix.tarantool }}
+        if: matrix.tarantool != 'master'
         uses: tarantool/setup-tarantool@v2
         with:
           tarantool-version: ${{ matrix.tarantool }}
+
+      - name: Get Tarantool master latest commit
+        if: matrix.tarantool == 'master'
+        run: |
+          commit_hash=$(git ls-remote https://github.com/tarantool/tarantool.git --branch master | head -c 8)
+          echo "LATEST_COMMIT=${commit_hash}" >> $GITHUB_ENV
+        shell: bash
+
+      - name: Cache Tarantool master
+        if: matrix.tarantool == 'master'
+        id: cache-latest
+        uses: actions/cache@v3
+        with:
+          path: |
+            ${{ github.workspace }}/bin
+            ${{ github.workspace }}/include
+          key: cache-latest-${{ env.LATEST_COMMIT }}
+
+      - name: Setup Tarantool master
+        if: matrix.tarantool == 'master' && steps.cache-latest.outputs.cache-hit != 'true'
+        run: |
+          tt install tarantool master
+
+      - name: Add Tarantool master to PATH
+        if: matrix.tarantool == 'master'
+        run: echo "${GITHUB_WORKSPACE}/bin" >> $GITHUB_PATH
 
       - name: Setup Python for tests
         uses: actions/setup-python@v4
@@ -86,8 +121,6 @@ jobs:
 
       - name: Install the crud module for testing purposes
         run: |
-          curl -L https://tarantool.io/release/2/installer.sh | bash
-          sudo apt install -y tt
           tt rocks install crud
 
       - name: Run tests
--- a/tarantool/connection.py
+++ b/tarantool/connection.py
@@ -4,6 +4,7 @@ This module provides API for interaction with a Tarantool server.
 # pylint: disable=too-many-lines,duplicate-code
 
 import os
+import select
 import time
 import errno
 from enum import Enum
@@ -51,6 +52,9 @@ from tarantool.const import (
     RECONNECT_DELAY,
     DEFAULT_TRANSPORT,
     SSL_TRANSPORT,
+    DEFAULT_HOST,
+    DEFAULT_PORT,
+    DEFAULT_SOCKET_FD,
     DEFAULT_SSL_KEY_FILE,
     DEFAULT_SSL_CERT_FILE,
     DEFAULT_SSL_CA_FILE,
@@ -594,7 +598,10 @@ class Connection(ConnectionInterface):
     :value: :exc:`~tarantool.error.CrudModuleError`
     """
 
-    def __init__(self, host, port,
+    def __init__(self,
+                 host=DEFAULT_HOST,
+                 port=DEFAULT_PORT,
+                 socket_fd=DEFAULT_SOCKET_FD,
                  user=None,
                  password=None,
                  socket_timeout=SOCKET_TIMEOUT,
@@ -623,8 +630,11 @@ class Connection(ConnectionInterface):
             Unix sockets.
         :type host: :obj:`str` or :obj:`None`
 
-        :param port: Server port or Unix socket path.
-        :type port: :obj:`int` or :obj:`str`
+        :param port: Server port, or Unix socket path.
+        :type port: :obj:`int` or :obj:`str` or :obj:`None`
+
+        :param socket_fd: socket fd number.
+        :type socket_fd: :obj:`int` or :obj:`None`
 
         :param user: User name for authentication on the Tarantool
             server.
@@ -804,6 +814,18 @@ class Connection(ConnectionInterface):
         """
         # pylint: disable=too-many-arguments,too-many-locals,too-many-statements
 
+        if host is None and port is None and socket_fd is None:
+            raise ConfigurationError("need to specify host/port, "
+                                     "port (in case of Unix sockets) "
+                                     "or socket_fd")
+
+        if socket_fd is not None and (host is not None or port is not None):
+            raise ConfigurationError("specifying both socket_fd and host/port is not allowed")
+
+        if host is not None and port is None:
+            raise ConfigurationError("when specifying host, "
+                                     "it is also necessary to specify port")
+
         if msgpack.version >= (1, 0, 0) and encoding not in (None, 'utf-8'):
             raise ConfigurationError("msgpack>=1.0.0 only supports None and "
                                      + "'utf-8' encoding option values")
@@ -820,6 +842,7 @@ class Connection(ConnectionInterface):
         recv.restype = ctypes.c_int
         self.host = host
         self.port = port
+        self.socket_fd = socket_fd
         self.user = user
         self.password = password
         self.socket_timeout = socket_timeout
@@ -897,10 +920,37 @@ class Connection(ConnectionInterface):
         :meta private:
         """
 
-        if self.host is None:
-            self.connect_unix()
-        else:
+        if self.socket_fd is not None:
+            self.connect_socket_fd()
+        elif self.host is not None:
             self.connect_tcp()
+        else:
+            self.connect_unix()
+
+    def connect_socket_fd(self):
+        """
+        Establish a connection using an existing socket fd.
+
+            +---------------------+--------------------------+-------------------------+
+            | socket_fd / timeout |         >= 0             |         `None`          |
+            +=====================+==========================+=========================+
+            |      blocking       | Set non-blocking socket  | Don't change, `select`  |
+            |                     | lib call `select`        | isn't needed            |
+            +---------------------+--------------------------+-------------------------+
+            |     non-blocking    | Don't change, socket lib | Don't change, call      |
+            |                     | call `select`            | `select` ourselves      |
+            +---------------------+--------------------------+-------------------------+
+
+        :meta private:
+        """
+
+        self.connected = True
+        if self._socket:
+            self._socket.close()
+
+        self._socket = socket.socket(fileno=self.socket_fd)
+        if self.socket_timeout is not None:
+            self._socket.settimeout(self.socket_timeout)
 
     def connect_tcp(self):
         """
@@ -1124,6 +1174,11 @@ class Connection(ConnectionInterface):
         while to_read > 0:
             try:
                 tmp = self._socket.recv(to_read)
+            except BlockingIOError:
+                ready, _, _ = select.select([self._socket.fileno()], [], [], self.socket_timeout)
+                if not ready:
+                    raise NetworkError(TimeoutError())  # pylint: disable=raise-missing-from
+                continue
             except OverflowError as exc:
                 self._socket.close()
                 err = socket.error(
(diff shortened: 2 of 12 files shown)


tarantool/connection.py
```diff
@@ -897,10 +897,33 @@ def connect_basic(self):
...
+        :meta private:
+        """
+
+        try:
+            # If old socket already exists - close it and re-create.
+            self.connected = True
+            if self._socket:
+                self._socket.close()
+            self._socket = socket.socket(fileno=self.port)
+            self._socket.settimeout(self.socket_timeout)
```
oleg-jukovec, line 923: Does the `fd` socket from tarantool is blocking?

askalt: As I understood, there is no difference. If `self.socket_timeout` is None, the socket is put in blocking mode. If is not None, will be used non-blocking socket (similar to other connections).

oleg-jukovec: I suppose it is not expected that the fd socket may change the type from non-blocking to blocking.

askalt:
There will be one more test (with Tarantool made sockets and child process), but in general it is already ready for the review.

test/suites/test_socket_fd.py
```diff
@@ -0,0 +1,107 @@
...
+
+            socket.tcp_server('0.0.0.0', %d, function(s)
+                box.session.new({
+                    type = 'binary',
+                    fd = s:fd(),
+                    user = 'test',
+                })
+                s:detach()
+            end)
+
```
oleg-jukovec, line 48: What does it for the tests?

test/suites/test_socket_fd.py
```diff
@@ -0,0 +1,107 @@
...
+        assert_admin_success(resp)
+
+    @skip_or_run_box_session_new_tests
+    def setUp(self):
+        # Prevent a remote tarantool from clean our session.
+        if self.srv.is_started():
+            self.srv.touch_lock()
+
+    def _get_tt_sock(self):
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
oleg-jukovec, line 60: The socket in the blocking mode?

test/suites/test_socket_fd.py
```diff
@@ -0,0 +1,107 @@
...
+            local socket = require('socket')
+
+            box.cfg{}
+            box.schema.user.create('test', {password = 'test', if_not_exists = true})
+            box.schema.user.grant('test', 'read,write,execute,create', 'universe',
+                                   nil, {if_not_exists = true})
+            box.schema.user.grant('guest', 'execute', 'universe',
+                                   nil, {if_not_exists = true})
+
+            socket.tcp_server('0.0.0.0', %d, function(s)
```
DifferentialOrange, line 40: The main usecase of box.session.new is plain UNIX sockets. So if we want to test full integration cycle, the scenario should be as follows.

Test suite starts a tarantool instance.
Tarantool instance creates a new session.
It also starts some small Python script with popen+inherit fds, which uses connect over fd.
Script triggers some small change that we can verify (for example, inserts a record.)
Test suite verify that the change had been applied.
(The last check is likely to be wrapped in a retry cycle.)

I'm not sure how complicated is to write such a test. The main questions are "how Tarantool instances started from unittest would resolve paths to the script" and "will script started with Tarantool succeed in package path resolve". It is likely that you will be required to write venv and no-venv version of the test.

tarantool/mesh_connection.py
```diff
@@ -298,7 +298,7 @@ def __init__(self, host=None, port=None,
             :paramref:`~tarantool.MeshConnection.params.addrs` list.
 
         :param port: Refer to
-            :paramref:`~tarantool.Connection.params.host`.
+            :paramref:`~tarantool.Connection.params.port`.
```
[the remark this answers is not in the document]
askalt: Changed.

tarantool/connection.py
```diff
@@ -897,10 +897,36 @@ def connect_basic(self):
...
         """
 
-        if self.host is None:
-            self.connect_unix()
-        else:
+        if self.host is not None:
             self.connect_tcp()
+            return
+        if isinstance(self.port, int):
+            self.connect_socket_fd()
```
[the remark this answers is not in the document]
askalt: Changed.

test/suites/test_socket_fd.py
askalt: Yes, there we test blocking sockets.

.github/workflows/testing.yml
```diff
@@ -27,6 +27,7 @@ jobs:
           - '2.8'
           - '2.10'
```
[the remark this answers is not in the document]
askalt: Changed.

tarantool/connection.py
```diff
@@ -897,10 +897,36 @@ def connect_basic(self):
...
+            return
+        self.connect_unix()
```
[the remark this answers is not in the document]
askalt: I add some changes to this function, so now it have different from `connect_unix()` logic.

tarantool/connection.py
```diff
@@ -1124,6 +1140,9 @@ def _recv(self, to_read):
         while to_read > 0:
             try:
                 tmp = self._socket.recv(to_read)
+            except BlockingIOError:
+                select.select([self._socket.fileno()], [], [])
```
oleg-jukovec, line 1144: A `timeout` should be here.

tarantool/connection.py
```diff
@@ -1163,6 +1182,39 @@ def _read_response(self):
...
+                if sent == 0:
+                    err = socket.error(
+                        errno.ECONNRESET,
+                        "Lost connection to server during query"
+                    )
+                    raise NetworkError(err)
+                total_sent += sent
+            except BlockingIOError as exc:
+                total_sent += exc.characters_written
+                select.select([], [self._socket.fileno()], [])
```
oleg-jukovec, line 1209: + `timeout`

oleg-jukovec (changes requested):
1. `select` must have respect the timeout setting for non-blocking sockets.
2. We need to think what we could to do if fd the socket is blocking and the timeout is set. The only way I see is to turn the socket into non-blocking mode.

tarantool/connection.py
askalt: Fixed, if select returns an empty list raise exception.

tarantool/connection.py
oleg-jukovec: The behavior is the same as python does?

tarantool/connection.py
```diff
@@ -897,10 +898,36 @@ def connect_basic(self):
...
+        elif isinstance(self.port, int):
+            self.connect_socket_fd()
+        else:
+            self.connect_unix()
+
+    def connect_socket_fd(self):
+        """
+        Establish a connection using an existing socket fd.
+
+        socket_fd / timeout|         > 0             |         None
```
oleg-jukovec, line 912:
```suggestion
        socket_fd / timeout|         >= 0             |         None
```

askalt: Fixed.

tarantool/connection.py
```diff
@@ -623,7 +624,7 @@ def __init__(self, host, port,
             Unix sockets.
         :type host: :obj:`str` or :obj:`None`
 
-        :param port: Server port or Unix socket path.
+        :param port: Server port, socket fd or Unix socket path.
```
oleg-jukovec, line 627: Maybe it would be clearer to make a separate parameter?

oleg-jukovec (approved):
Thank you for the patch!

tarantool/connection.py
askalt: Yes, I guess. From the `settimeout` method of socket library:

> If a non-zero value is given, subsequent socket operations will raise a [timeout](https://docs.python.org/3/library/socket.html#socket.timeout) exception if the timeout period value has elapsed before the operation has completed.

oleg-jukovec: > Yes, I guess. From the `settimeout` method of socket library:
> 
> > If a non-zero value is given, subsequent socket operations will raise a [timeout](https://docs.python.org/3/library/socket.html#socket.timeout) exception if the timeout period value has elapsed before the operation has completed.

exception socket.timeout

    A deprecated alias of [TimeoutError](https://docs.python.org/3/library/exceptions.html#TimeoutError).

It could be better to use TimeoutError.

test/suites/test_socket_fd.py
DifferentialOrange: That's a nice update.

But I still wish to see unix socket session test.

DifferentialOrange: Oh, sorry, now I see

test/suites/test_socket_fd.py
```diff
@@ -0,0 +1,175 @@
...
+                return 1, err
+            end
+
+            ph:wait()
+
+            local status_code = ph:info().status.exit_code
+            local stderr = ph:read({stderr=true}):rstrip()
+            return status_code, stderr
+        """ % (cwd, python_exe, side_script_path))
+        if err is not None:
```
DifferentialOrange, line 167: You can use the third argument: `assertEqual(first, second, msg=None)`

askalt: Changed.

tarantool/connection.py
askalt: Changed to `TimeoutError`.

tarantool/connection.py
```diff
@@ -897,10 +907,35 @@ def connect_basic(self):
...
+        Establish a connection using an existing socket fd.
+
+        socket_fd / timeout|         >= 0            |         None
+        -------------------|-------------------------|-------------------------
+              blocking     | set non-blocking        |  don't change
+                           | socket lib call `select`| `select` isn't needed
+        -------------------|-------------------------|-------------------------
+            non-blocking   | don't change            |  don't change
+                           | socket lib call `select`|  call `select` ourselves
+        -------------------|-------------------------|-------------------------
```
DifferentialOrange, line 928: Current table formatting is not good for sphinx. You can test it by yourself by removing `:meta private:` and building the docs. Use this approach instead. (AFAIR, additional indentation is also required.)

```suggestion
            +---------------------+--------------------------+-------------------------+
            | socket_fd / timeout |         >= 0             |         `None`          |
            +=====================+==========================+=========================+
            |      blocking       | Set non-blocking socket  | Don't change, `select`  |
            |                     | lib call `select`        | isn't needed            |
            +---------------------+--------------------------+-------------------------+
            |     non-blocking    | Don't change, socket lib | Don't change, call      |
            |                     | call `select`            | `select` ourselves      |
            +---------------------+--------------------------+-------------------------+
```

tarantool/connection.py
```diff
@@ -1163,6 +1203,40 @@ def _read_response(self):
         # Read the packet
         return self._recv(length)
 
+    def _sendall(self, bytes_to_send):
+        """
+        Sends bytes to the transport (socket).
+
+        :param bytes_to_send: message to send.
```
DifferentialOrange, line 1210:
```suggestion
        :param bytes_to_send: Message to send.
```

tarantool/connection.py
DifferentialOrange: Well, we had decided to add a separate parameter

tarantool/connection.py
askalt: Fixed.

tarantool/connection.py
askalt: Fixed.

tarantool/connection.py
```diff
@@ -1163,6 +1203,40 @@ def _read_response(self):
...
+        :type bytes_to_send: :obj:`bytes`
+
+        :raise: :exc:`~tarantool.error.NetworkError`
+
+        :meta: private:
+        """
```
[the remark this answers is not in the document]
askalt: Added.

tarantool/connection.py
```diff
@@ -897,10 +907,35 @@ def connect_basic(self):
...
+        -------------------|-------------------------|-------------------------
+              blocking     | set non-blocking        |  don't change
+                           | socket lib call `select`| `select` isn't needed
+        -------------------|-------------------------|-------------------------
+            non-blocking   | don't change            |  don't change
+                           | socket lib call `select`|  call `select` ourselves
+        -------------------|-------------------------|-------------------------
+
+        :meta private:
+        """
```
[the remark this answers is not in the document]
askalt: Added.

test/suites/test_socket_fd.py
```diff
@@ -0,0 +1,184 @@
...
+            conn.close()
+
+    @unittest.skipIf(sys.platform.startswith("win"),
+                     "Skip on Windows since it uses remote server")
+    def test_04_tarantool_made_socket(self):
+        python_exe = find_python()
+        cwd = os.getcwd()
+        side_script_path = os.path.join(cwd, "test", "suites", "sidecar.py")
+
+        # pylint:disable=C0209
```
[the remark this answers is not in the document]
askalt: Changed.

test/suites/test_socket_fd.py
```diff
@@ -0,0 +1,184 @@
...
+        print(' SOCKET FD '.center(70, '='), file=sys.stderr)
+        print('-' * 70, file=sys.stderr)
+
+        cls.srv = TarantoolServer()
+        cls.srv.script = 'test/suites/box.lua'
+        cls.srv.start()
+        cls.tcp_port = find_port()
+
+        # Start tcp server to test work with blocking sockets.
+        # pylint:disable=C0209
```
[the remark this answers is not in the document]
askalt: Changed.

test/suites/test_socket_fd.py
```diff
@@ -0,0 +1,184 @@
...
+        sock.connect((self.srv.host, self.tcp_port))
+        return sock
+
+    def test_01_incorrect_params(self):
+        params = [
+            {"host": "123", "socket_fd": 3},
+            {"port": 14, "socket_fd": 3},
+        ]
+        for args in params:
+            with self.assertRaises(tarantool.Error):
```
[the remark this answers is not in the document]
askalt: Test rewritten in such style.

tarantool/mesh_connection.py
```diff
@@ -127,9 +134,6 @@ def format_error(address, err):
     if not isinstance(address, dict):
```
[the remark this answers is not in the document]
askalt: Checking for the absence of `host/port` when `socket_fd` is set has been added. "some parameter is not None" check is already there.

tarantool/const.py
```diff
@@ -103,6 +103,12 @@
 IPROTO_FEATURE_SPACE_AND_INDEX_NAMES = 5
 IPROTO_FEATURE_WATCH_ONCE = 6
 
+# Default value for host.
```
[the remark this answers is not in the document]
askalt: These variables useful [here](https://github.com/tarantool/tarantool-python/blob/master/tarantool/mesh_connection.py#L93-L94) and [here](https://github.com/tarantool/tarantool-python/blob/master/tarantool/mesh_connection.py#L137-L139). 
Changed to be consistent.

tarantool/connection.py
```diff
@@ -1163,6 +1203,40 @@ def _read_response(self):
...
+    def _sendall(self, bytes_to_send):
+        """
+        Sends bytes to the transport (socket).
+
+        :param bytes_to_send: message to send.
+        :type bytes_to_send: :obj:`bytes`
+
+        :raise: :exc:`~tarantool.error.NetworkError`
+
+        :meta: private:
```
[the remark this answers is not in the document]
askalt: Changed.

tarantool/connection.py
```diff
@@ -803,6 +810,8 @@ def __init__(self, host, port,
         .. _mp_array: https://github.com/msgpack/msgpack/blob/master/spec.md#array-format-family
         """
         # pylint: disable=too-many-arguments,too-many-locals,too-many-statements
+        if socket_fd is not None and (host is not None or port is not None):
```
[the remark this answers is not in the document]
askalt: Changed, test added.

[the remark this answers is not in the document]
askalt: Added.

tarantool/connection.py
```diff
@@ -804,6 +814,18 @@ def __init__(self, host, port,
         """
         # pylint: disable=too-many-arguments,too-many-locals,too-many-statements
 
+        if host is None and port is None and socket_fd is None:
+            raise ConfigurationError("need to specify both host and port, "
+                                     "port only in the case on Unix sockets, "
+                                     "or socket_fd in the case of socket fd")
```
DifferentialOrange, line 820:
```suggestion
raise ConfigurationError("need to specify host/port, "
                                     "port (in case of Unix sockets)"
                                     "or socket_fd")
```

DifferentialOrange (approved):
Sorry for the long delay, I was a bit busy. Let's fix the last description and I'll merge it.

DifferentialOrange (approved):

PR #320, opened 2024-03-29
test: use coverage instead of equality to test error payload

nshy:
We are going to add missing 'user' payload field for ACCESS_DENIED error which will break current tests. Let rewrite tests to allow adding new payload fields for this error.

Need for https://github.com/tarantool/tarantool/issues/9108


DifferentialOrange (approved):
Thanks for the fixes!

PR #323, opened 2024-05-24
ci: bump ubuntu runner versions

DifferentialOrange:
This patch does not update dependency for reusable_testing since this script is a part of DevX team responsibility and needs a separate check.


DifferentialOrange:
Let's stick to 20.04 for a bit more until we decide to drop 3.6 support

PR #343, opened 2026-07-20
connection_pool: recover instances after a cluster restart

### LINKED ISSUE #328: ConnectionPool stuck on error after cluster restart
Ticket is based on a customer report.

Let ConnectionPool be connected to a cluster. After the whole cluster is restarted, it stucks on `Can't find healthy rw instance in pool` due to some `tarantool.error.DatabaseError: (116, "Instance bootstrap hasn't finished yet")` internal errors.

Mockird31:
Treat any instance error as an unhealthy state instead of an unhandled exception. An instance which has not finished its bootstrap yet replies with an error to `box.info`, and the `Response` constructor raises it as a plain DatabaseError. `_get_new_state()` caught only a NetworkError, so the error escaped the background refresh loop and killed its thread.

Close the socket if a handshake fails: `is_closed()` only checks the socket, so such a connection was reported as open and was never authenticated again.

Closes #328


oleg-jukovec:
Should be `Closes #328`

oleg-jukovec (approved):

